PR #17, opened 2014-11-24
Use TextNum analyzer by default

dantleech:
This PR switches to using the TextNumb analyzer by default for Zend Lucene search.

I am not sure this is desired, the current analyzer works fine for numbers within strings, but not for numbers only. i.e.
- `Some text 100`: OK
- `100` NOT OK

This PR allows the later.

Not really to keen on merging this just yet - I would rather create a PR to allow a choice of analyzers via. configuration.


wachterjohannes:
@dantleech we should finish this PR (:

PR #18, opened 2015-01-27
Feature/rest api

dantleech:
This PR adds a Doctrine ORM subscriber, REST API and ExpressionLanaugage and *numerous other* features facilitating the creation of an independent back office search.

- Added Doctrine ORM subscriber will automatically index/deindex mapped ORM documents.
- `massive:search:index:rebuild` command which rebuilds both the ORM and third-party (i.e. Sulu) implementations)
- <strike>Introduction of a REST API to provide a simple controller to perform a search query using parameters `query`, `index` and `locale` and receieve a JSON response.</strike>
- Introduction of multiple search *contexts* for each mapped object, possiblity to index things differently for back or front office searches.
- Added behat tests
- Category support

- Docs in separate PR: https://github.com/massiveart/MassiveSearchBundle/pull/20


Search/EventSubscriber/DoctrineOrmSubscriber.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
```
dantleech: Missing header

Search/Exception/MetadataNotFoundException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
```
dantleech: Missing header

Tests/Resources/app/config/massivesearchbundle.yml
```diff
@@ -1,2 +1,3 @@
 massive_search:
+    adapter: test
```
dantleech: Use the in-memory test adapter by default for the Functional tests.

Tests/Unit/Search/EventSubscriber/DoctrineOrmSubscriberTest.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
```
dantleech: Header

Tests/Unit/Search/SearchManagerTest.php
```diff
@@ -87,7 +88,7 @@ public function testIndexNonObject()
     }
 
     /**
-     * @expectedException \RuntimeException
+     * @expectedException Massive\Bundle\SearchBundle\Search\Exception\MetadataNotFoundException
```
dantleech: do not expand, use

composer.json
```diff
@@ -18,8 +18,11 @@
         "elasticsearch/elasticsearch": "~1.3",
         "symfony-cmf/testing": "dev-master",
         "symfony/filesystem": "~2.4",
+        "symfony/console": "~2.5",
         "phpspec/prophecy-phpunit": "~1.0",
-        "matthiasnoback/symfony-dependency-injection-test": "0.7.1"
+        "matthiasnoback/symfony-dependency-injection-test": "0.7.1",
+        "doctrine/doctrine-bundle": "~1.3",
```
dantleech: DoctrineBundle is not required here... or ...coming to think of it I am missing the tag for the ORM subscriber.

Controller/RestController.php
```diff
@@ -0,0 +1,48 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Controller;
+
+use Massive\Bundle\SearchBundle\Search\SearchManagerInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+
+class RestController
```
dantleech: Missing doc

Resources/config/doctrine_orm.xml
```diff
@@ -0,0 +1,16 @@
...
+        <parameter key="massive_search.search.event_subscriber.doctrine_orm.class">Massive\Bundle\SearchBundle\Search\EventSubscriber\DoctrineOrmSubscriber</parameter>
+    </parameters>
+
+    <services>
+        <service id="massive_search.search.event_subscriber.doctrine_orm" class="%massive_search.search.event_subscriber.doctrine_orm.class%">
+            <argument type="service" id="massive_search.search_manager" />
+        </service>
+    </services>
+</container>
+
```
dantleech: Remove empty line

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,216 @@
...
+     * @param mixed $field
+     */
+    private function getValue($object, $field)
+    {
+        switch (get_class($field)) {
+            case 'Massive\Bundle\SearchBundle\Search\Metadata\Property':
+                return $this->getPropertyValue($object, $field);
+            case 'Massive\Bundle\SearchBundle\Search\Metadata\Expression':
+                return $this->getExpressionValue($object, $field);
+
```
dantleech: EL

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,216 @@
...
+        return $this->accessor->getValue($object, $field->getProperty());
+    }
+
+    /**
+     * Evaluate an expression (ExpressionLanguage)
+     *
+     * @param mixed $object
+     * @param mixed $field
+     */
+    private function getExpressionValue($object, $field)
```
dantleech: type hint

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,216 @@
...
+        ));
+    }
+
+    /**
+     * Evaluate a property (using PropertyAccess)
+     *
+     * @param mixed $object
+     * @param mixed $field
+     */
+    private function getPropertyValue($object, $field)
```
dantleech: type hint

Tests/Functional/Controller/RestControllerTest.php
```diff
@@ -0,0 +1,81 @@
...
+                        'document' => array(
+                            'id' => 6,
+                            'title' => 'Product X',
+                            'description' => 'To be or not to be, that is the question',
+                            'class' => 'Massive\\Bundle\\SearchBundle\\Tests\\Resources\\TestBundle\\Entity\\Product',
+                            'url' => '/foobar',
+                            'image_url' => null,
+                            'locale' => 'fr',
+                        ),
+                        'score' => -1,
```
dantleech: @danrot @chirimoya @wachterjohannes this would be the JSON response used for the admin search.

Tests/Resources/TestBundle/Resources/config/massive-search/Product.xml
```diff
@@ -1,11 +1,13 @@
...
-        <urlField name="url" />
-        <titleField name="title" />
-        <descriptionField name="body" />
+        <index name="product"/>
+        <id property="id"/>
+        <url property="url" />
+        <title property="title" />
+        <locale property="locale" />
+        <description property="body" />
+        <image property="image" />
```
dantleech: bc break ..

Tests/Unit/Search/ObjectToDocumentConverterTest.php
```diff
@@ -0,0 +1,92 @@
...
+use Massive\Bundle\SearchBundle\Search\Metadata\IndexMetadata;
+use Massive\Bundle\SearchBundle\Tests\Resources\TestBundle\Entity\Product;
+use Prophecy\PhpUnit\ProphecyTestCase;
+use Massive\Bundle\SearchBundle\Search\Metadata\Property;
+use Massive\Bundle\SearchBundle\Search\ObjectToDocumentConverter;
+use Massive\Bundle\SearchBundle\Search\Factory;
+use Massive\Bundle\SearchBundle\Search\Metadata\Expression;
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+
+class ObjectToDocumentConverterTest extends ProphecyTestCase
```
dantleech: Docs for this class

Tests/Unit/Search/ObjectToDocumentConverterTest.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    protected $indexMetadata;
```
dantleech: Add properties

dantleech:
todo:

- Context: Different URLs (and possibly images, body, title etc) are required for backend and frontend, should add a context `context=backend` for the mappings.

Tests/Resources/TestBundle/Resources/config/massive-search/Product.xml
dantleech: also should add some expressions here, `<url expr="..."/>` as I think this is used for the XML driver test.

Search/Metadata/Property.php
```diff
@@ -0,0 +1,32 @@
...
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search\Metadata;
+
+/**
+ * Simple value object for representing propertys
```
dantleech: properties

Search/Metadata/Property.php
```diff
@@ -0,0 +1,32 @@
...
+    {
+        $this->property = $property;
+    }
+
+    public function getProperty()
+    {
+        return $this->property;
+    }
+}
+
```
dantleech: Docs + spaces

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,216 @@
...
+     * @var PropertyAccess
+     */
+    private $accesor;
+
+    /**
+     * @var Factory
+     */
+    private $factory;
+
+    public function __construct(Factory $factory, ExpressionLanguage $expressionLanguage)
```
dantleech: Doc

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,216 @@
...
+
+                foreach ($childObjects as $i => $childObject) {
+                    $this->populateDocument(
+                        $document,
+                        $childObject,
+                        $mapping['mapping']->getFieldMapping(),
+                        $prefix . $field . $i
+                    );
+                }
+            } else {
```
dantleech: `else` are not required. use `continue`

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ CHANGELOG
 dev-master
 ----------
 
+- [TestAdapter] Supports indexes
+- [Mapping] Expression language
+- [REST] REST API
+- [Persistence] Doctrine ORM event subscriber
```
dantleech: Adds purge and rebuild support

README.rst
```diff
@@ -122,10 +123,123 @@ through *mapping*. Currently only **XML mapping** supported:
...
+            <title property="title" />
+            <url expr="'/path/to/' ~ object.id" />
+            <description property="body" />
+            <fields>
+                <field name="title" type="string" />
+                <field name="body" type="string" />
+            </fields>
+
+            <context name="admin">
+                <url exp="'/admin/edit/article/' ~ object.id" />
```
dantleech: Contexts are only useful when you use a different index. Update this example.

Resources/config/routing.yml
```diff
@@ -0,0 +1,5 @@
+massive_search:
+    path: /search
+    defaults:
+        _controller: "massive_search.controller.rest:searchAction"
+
```
dantleech: Remove EL

Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -185,12 +186,42 @@ public function search(SearchQuery $searchQuery)
...
+        foreach ($indexes as $indexName => $index) {
+            foreach ($index as $field => $value) {
+                $status['idx:' . $indexName . '.' . $field] = substr(trim(json_encode($value)), 0, 100);
+            }
         }
 
         return $status;
     }
+
+    public function purge($indexName)
```
dantleech: Docs for this class.

Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -185,12 +186,42 @@ public function search(SearchQuery $searchQuery)
...
+    {
+        $indexes = $this->listIndexes();
+
+        foreach (array_keys($indexes) as $realIndexName) {
+            $isLocalizedVersion = $this->localizationStrategy->isIndexVariantOf($indexName, $realIndexName);
```
dantleech: `$isVariant`

Search/Adapter/TestAdapter.php
```diff
@@ -33,18 +33,27 @@ public function __construct(Factory $factory)
...
-                unset($this->documents[$i]);
+                unset($this->documents[$indexName][$i]);
             }
         }
 
-        $this->documents = array_values($this->documents);
+        $this->documents[$indexName] = array_values($this->documents[$indexName]);
+    }
+
+    public function purge($indexName)
```
dantleech: Docs for this class

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -285,4 +287,35 @@ private function getIndex($indexPath, $create = false)
...
+            if (!$isLocalizedVersion) {
+                continue;
+            }
+
+            $indexPath = $this->basePath . '/' . $index;
+            $fs->remove($indexPath);
+        }
+    }
+
+    private function listIndexes()
```
dantleech: Docs for this class

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -202,6 +203,7 @@ public function getStatus()
             $files = $indexFinder->files()->name('*')->depth('== 0')->in($indexDir->getPathname());
             $indexName = basename($indexDir);
 
+
```
dantleech: Remove EL

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -285,4 +287,35 @@ private function getIndex($indexPath, $create = false)
...
     }
+
+    public function purge($indexName)
+    {
+        $indexPath = $this->getIndexPath($indexName);
+        $indexes= $this->listIndexes();
+
+        foreach ($indexes as $index) {
+            $fs = new Filesystem();
+            $isLocalizedVersion = $this->localizationStrategy->isIndexVariantOf($indexName, $index);
```
dantleech: `$isVariant`

Search/Event/IndexRebuildEvent.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return $this->purge;
+    }
+
+    public function getOutput() 
+    {
+        return $this->output;
+    }
+}
+
```
dantleech: Remove EL

Search/Event/IndexRebuildEvent.php
```diff
@@ -0,0 +1,47 @@
...
+     * @param OutputInterface $output 
+     */
+    public function __construct($filter, $purge, OutputInterface $output = null)
+    {
+        $this->filter = $filter;
+        $this->purge = $purge;
+        $this->output = $output ? : new NullOutput();
+    }
+
+    public function getFilter()
```
dantleech: Docs

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
```diff
@@ -0,0 +1,136 @@
...
+use Doctrine\ORM\Mapping\ClassMetadataFactory;
+use Massive\Bundle\SearchBundle\Search\SearchEvents;
+use Doctrine\Common\Persistence\ObjectManager;
+use Massive\Bundle\SearchBundle\Search\Event\IndexRebuildEvent;
+use Metadata\MetadataFactory;
+use Massive\Bundle\SearchBundle\Search\SearchManager;
+use Symfony\Component\Console\Output\OutputInterface;
+use Metadata\ClassMetadata as SearchMetadata;
+use Doctrine\Common\Persistence\Mapping\ClassMetadata as OrmMetadata;
+use Massive\Bundle\SearchBundle\Search\Metadata\IndexMetadata;
```
dantleech: Are all of these USE required? Run cs fixer.

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
```diff
@@ -0,0 +1,136 @@
...
+    {
+        $output = $event->getOutput();
+        $filter = $event->getFilter();
+        $purge = $event->getPurge();
+
+        $metadataFactory = $this->objectManager->getMetadataFactory();
+        $metadata = $metadataFactory->getAllMetadata();
+
+        foreach ($metadata as $class) {
+            if ($filter && !preg_match('{' . $filter . '}', $class->name)) {
```
dantleech: Document how this should work in the interface.

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
```diff
@@ -0,0 +1,136 @@
...
+     * {@inheritDoc}
+     */
+    public static function getSubscribedEvents()
+    {
+        return array(
+            SearchEvents::INDEX_REBUILD => 'rebuildIndex'
+        );
+    }
+
+    public function rebuildIndex(IndexRebuildEvent $event)
```
dantleech: Docs for this class.

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
```diff
@@ -0,0 +1,136 @@
...
+
+            $output->writeln('<info>Purging index</info>: ' . $indexName);
+            $this->searchManager->purge($indexName);
+            $this->purged[$indexName] = true;
+        }
+    }
+
+    private function rebuildClass(OutputInterface $output, OrmMetadata $ormMetadata)
+    {
+        $output->write('<comment>Rebuilding</comment>: ' . $ormMetadata->getName());
```
dantleech: It is maybe not the best strategy to have the listeners write directly to the `OutputInterface`. It would be better for the Command to be responsible for the display of data .. but for now.

dantleech:
Added behat tests for search adapters and removed old integration tests. Need to also add some basic Web API tests and review and update the docs.

Behat/SearchManagerContext.php
```diff
@@ -0,0 +1,208 @@
...
+use Behat\Behat\Context\SnippetAcceptingContext;
+use Behat\Gherkin\Node\PyStringNode;
+use Behat\Symfony2Extension\Context\KernelAwareContext;
+use Symfony\Component\Filesystem\Filesystem;
+use Massive\Bundle\SearchBundle\Tests\Resources\app\AppKernel;
+use Symfony\Component\HttpKernel\KernelInterface;
+use PHPUnit_Framework_Assert as Assert;
+use Massive\Bundle\SearchBundle\Search\SearchManager;
+
+class SearchManagerContext implements SnippetAcceptingContext, KernelAwareContext
```
dantleech: Doc

.gitignore
```diff
@@ -1,6 +1,9 @@
 Tests/Resources/app/cache
 Tests/Resources/app/logs
 Tests/Resources/app/data
+Tests/Resources/TestBundle/Resources/config/massive-search/
+Tests/Resources/TestBundle/Entity
+Tests/Resources/app/config/_behat_massivesearch.yml
```
dantleech: Remove

Behat/SearchManagerContext.php
```diff
@@ -0,0 +1,208 @@
...
+            $this->kernel->getContainer()->get('event_dispatcher'),
+            $this->kernel->getContainer()->get('massive_search.localization_strategy')
+        );
+    }
+
+    protected function pause()
+    {
+        // there is a timing issue, we need to pause for a while
+        // after flusing for subsequent requests to not fail.
+        usleep(50000);
```
dantleech: Move doc to block

Resources/config/metadata.xml
```diff
@@ -22,6 +23,12 @@
             <argument type="collection"/>
         </service>
 
+        <!-- Field evaluator -->
+        <service id="massive_search.metadata.field_evaluator" class="%massive_search.metadata.field_evaluator.class%">
+            <argument type="service" id="massive_search.expression_language" />
+        </service>
+
+
```
dantleech: EL

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -285,4 +260,32 @@ private function getIndex($indexPath, $create = false)
 
         return $index;
     }
+
+    public function purge($indexName)
```
dantleech: Dcos

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -285,4 +260,32 @@ private function getIndex($indexPath, $create = false)
...
+        return $names;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function flush(array $indexNames)
+    {
+    }
+
```
dantleech: EL

Search/Event/PreIndexEvent.php
```diff
@@ -23,34 +24,47 @@ class PreIndexEvent extends Event
...
+    private $fieldEvaluator;
 
     /**
      * @param $subject
      * @param Document $document
      * @param IndexMetadataInterface $metadata
      */
-    public function __construct($subject, Document $document, IndexMetadataInterface $metadata)
-    {
+    public function __construct(
```
dantleech: wrong dockblock

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
```diff
@@ -0,0 +1,134 @@
...
+
+        $objects = $this->objectManager->getRepository($ormMetadata->name)->findAll();
+
+        $count = 0;
+        foreach ($objects as $object) {
+            $this->searchManager->index($object);
+            $count++;
+        }
+        $output->writeln(sprintf(
```
dantleech: add docs

Search/Metadata/ClassMetadata.php
```diff
@@ -0,0 +1,51 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search\Metadata;
+
+use Metadata\ClassMetadata as BaseClassMetadata;
+
+/**
+ * Represents the
```
dantleech: finish that sentanence

Search/Metadata/ClassMetadata.php
```diff
@@ -0,0 +1,51 @@
...
+        $indexMetadata->setClassMetadata($this);
+        $this->indexMetadatas[$contextName] = $indexMetadata;
+    }
+
+    public function getIndexMetadatas()
+    {
+        return $this->indexMetadatas;
+    }
+
+    public function getName()
```
dantleech: docs

Search/Metadata/Driver/XmlDriver.php
```diff
@@ -59,30 +76,164 @@ public function loadMetadataFromFile(\ReflectionClass $class, $file)
...
+        }
+
+        return $this->getField($field);
+    }
+
+    private function getField(\SimpleXmlElement $field)
+    {
+        if (isset($field['expr']) && isset($field['property'])) {
+            throw new \InvalidArgumentException(sprintf(
+                '"expr" and "proprty" attributes are mutually exclusive in mapping for "%s"',
```
dantleech: porprty => property

Search/Metadata/IndexMetadata.php
```diff
@@ -200,4 +217,14 @@ public function setLocaleField($localeField)
     {
         $this->localeField = $localeField;
     }
+
+    public function getClassMetadata()
+    {
+        return $this->classMetadata;
+    }
+
+    public function setClassMetadata(ClassMetadata $classMetadata)
```
dantleech: docs

Behat/SearchManagerContext.php
```diff
@@ -0,0 +1,204 @@
...
+
+class SearchManagerContext implements SnippetAcceptingContext, KernelAwareContext
+{
+    private $adapterId;
+    private $kernel;
+    private $lastResult;
+    private $entityClasses;
+    private $entities = array();
+
+    public function __construct($adapterId)
```
dantleech: Docs etc.

Search/Metadata/Field/Property.php
```diff
@@ -0,0 +1,44 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search\Metadata\Field;
+
+use Massive\Bundle\SearchBundle\Search\Metadata\FieldInterface;
+
+/**
+ * Simple value object for representing propertys
```
dantleech: propertys => properties -- what does it use? PropertyAccessor

Search/Metadata/FieldEvaluator.php
```diff
@@ -0,0 +1,125 @@
...
+namespace Massive\Bundle\SearchBundle\Search\Metadata;
+
+use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
+use Symfony\Component\PropertyAccess\PropertyAccess;
+use Massive\Bundle\SearchBundle\Search\Metadata\Field\Property;
+use Massive\Bundle\SearchBundle\Search\Metadata\Field\Field;
+use Massive\Bundle\SearchBundle\Search\Metadata\Field\Expression;
+
+/**
+ * Evaluate the value of fields
```
dantleech: .. using expressionlangyuage

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,188 @@
...
+            }
+
+            if (!isset($mapping['type'])) {
+                throw new \RuntimeException(sprintf(
+                    'Mapping for "%s" does not have "type" key',
+                    $fieldName
+                ));
+            }
+
+            if ($mapping['type'] == 'complex') {
```
dantleech: Document this.

Search/SearchManager.php
```diff
@@ -93,10 +104,16 @@ public function getMetadata($object)
     public function deindex($object)
     {
         $metadata = $this->getMetadata($object);
-        $indexName = $metadata->getIndexName();
-        $document = $this->objectToDocument($metadata, $object);
 
-        $this->adapter->deindex($document, $indexName);
+        foreach ($metadata->getIndexMetadatas() as $indexMetadata) {
+            $this->indexesToFlush[$indexMetadata->getIndexName()] = true;
```
dantleech: encalsualte in method, e.g. `queueFlush($indexName)`

Search/SearchManager.php
```diff
@@ -178,132 +204,91 @@ public function getStatus()
...
-                            print_r($mapping, true)
-                        )
-                    );
-                }
+        $this->adapter->flush(array_keys($this->indexesToFlush));
+        $this->indexesToFlush = array();
+    }
 
-                $childObjects = $accessor->getValue($object, $field);
+    private function getExpandedIndexNamesFor($indexName, $locale = null)
```
dantleech: document this

Tests/Features/search_manager.feature
```diff
@@ -0,0 +1,121 @@
...
+
+    Scenario: Basic indexing
+        Given I index the following "Car" objects
+        """
+        [
+            { "id": 123, "url": "/url/to", "title": "My car", "body": "Hello", "image": "foo.jpg"},
+            { "id": 321, "url": "/url/to", "title": "My car", "body": "Hello", "image": "foo.jpg"}
+        ]
+        """
+        And I search for "My car"
```
dantleech: `When I search for "My car"`

dantleech: And also all the other things here.

Tests/Features/search_manager.feature
```diff
@@ -0,0 +1,121 @@
...
+            { "id": 123, "url": "/url/to", "title": "My car", "body": "Hello", "image": "foo.jpg"},
+            { "id": 321, "url": "/url/to", "title": "My car", "body": "Hello", "image": "foo.jpg"}
+        ]
+        """
+        And I purge the index "car"
+        And I search for "My car"
+        Then there should be 0 results
+
+    Scenario: Deindex
+        Given I index the following "Car" objects
```
dantleech: `Given the following "Car" objects exist`

Tests/Features/search_manager_localized.feature
```diff
@@ -0,0 +1,76 @@
...
+        Given I index the following "Product" objects
+        """
+        [
+            { "id": 1, "title": "Giraffe", "body": "Long neck", "date": "2015-01-01", "url": "http://foo", "locale": "fr", "image": "foo.png" },
+            { "id": 2, "title": "Lion", "body": "Big mane", "date": "2015-01-01", "url": "http://lion.com", "locale": "fr", "image": "foo.png" },
+            { "id": 4, "title": "Lion 2", "body": "Big mane", "date": "2015-01-01", "url": "http://lion.com", "locale": "fr", "image": "foo.png" },
+            { "id": 6, "title": "German Hyena", "body": "Laughs", "date": "2015-01-01", "url": "http://hyena.com", "locale": "de", "image": "foo.png" }
+        ]
+        """
+        And I search for "<animal>" in locale "<locale>"
```
dantleech: s/And/When

Tests/Functional/Controller/RestControllerTest.php
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+namespace Massive\Bundle\SearchBundle\Tests\Functional\Search;
+
+use Massive\Bundle\SearchBundle\Controller\RestController;
+use Massive\Bundle\SearchBundle\Tests\Functional\BaseTestCase;
+use Massive\Bundle\SearchBundle\Tests\Resources\TestBundle\Entity\Product;
+use Symfony\Component\HttpFoundation\Request;
+
+class RestControllerTest extends BaseTestCase
```
dantleech: Note this should one day be converted to a Behat test. But for now its fine.

Tests/Functional/Controller/RestControllerTest.php
```diff
@@ -0,0 +1,90 @@
...
+     * @var RestController
+     */
+    private $controller;
+
+    /**
+     * @var SearchManager
+     */
+    private $searchManager;
+
+    private $client;
```
dantleech: doc

Tests/Functional/Search/SearchManagerTest.php
```diff
@@ -8,27 +8,28 @@
...
  */
 
-namespace Massive\Bundle\SearchBundle\Tests\Functional;
+namespace Massive\Bundle\SearchBundle\Tests\Functional\Search;
 
 use Massive\Bundle\SearchBundle\Tests\Resources\TestBundle\Entity\Product;
 use Massive\Bundle\SearchBundle\Tests\Resources\TestBundle\EventSubscriber\TestSubscriber;
+use Massive\Bundle\SearchBundle\Tests\Functional\BaseTestCase;
 
 class SearchManagerTest extends BaseTestCase
```
dantleech: Remove this test and everything associated with it. Covered with Behat test.

Tests/Resources/app/config/massivesearchbundle.yml
```diff
@@ -1,2 +1,16 @@
+framework:
+    router:
+        resource: %kernel.root_dir%/config/routing.yml
+
+
 massive_search:
+
```
dantleech: EL

dantleech:
Change NS doinain from `massive.io` to `massiveart.com`

dantleech:
Build failure is an easy fix

Command/IndexRebuildCommand.php
```diff
@@ -0,0 +1,49 @@
...
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Massive\Bundle\SearchBundle\Search\Event\IndexRebuildEvent;
+use Massive\Bundle\SearchBundle\Search\SearchEvents;
+
+/**
+ * Comand to build (or rebuild) the search index
+ *
+ * @author Daniel Leech <daniel@dantleech.com>
```
danrot: Actually we agreed on omitting the author tag in all kind of documentation, since it will be wrong at some point anyway.

Command/QueryCommand.php
```diff
@@ -53,9 +53,9 @@ public function execute(InputInterface $input, OutputInterface $output)
         $locale = $input->getOption('locale');
 
         $searchManager = $this->getContainer()->get('massive_search.search_manager');
-        $res = $searchManager->createSearch($query)->index($index)->locale($locale)->execute();
+        $res = $searchManager->createSearch($query)->indexes($index)->locale($locale)->execute();
```
danrot: Is there a reason for changing this name?

Command/StatusCommand.php
```diff
@@ -10,11 +10,10 @@
...
 namespace Massive\Bundle\SearchBundle\Command;
 
-use Symfony\Component\Console\Command\Command;
-use Symfony\Component\Console\Helper\TableHelper;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Helper\Table;
 
 class StatusCommand extends ContainerAwareCommand
```
danrot: PHPDoc missing

Command/IndexRebuildCommand.php
dantleech: Force of habbit. As explained before I think it is a way of people signing their work, and ultimately claiming some responsiblitiy, it also gives people exposure to their peers. Its an incentive do good work imo.

Controller/RestController.php
```diff
@@ -0,0 +1,56 @@
...
+        $indexes = $request->query->get('indexes') ? : array();
+        $locale = $request->query->get('locale') ? : null;
+
+        $hits = $this->searchManager
+            ->createSearch($query)
+            ->locale($locale)
+            ->indexes($indexes)
+            ->execute();
+
+        return new JsonResponse($hits);
```
danrot: Why don't you use the FOSRestBundle here? Would be nice to be able to respond e.g. in XML as well.

DependencyInjection/MassiveSearchExtension.php
```diff
@@ -36,7 +36,19 @@ public function load(array $configs, ContainerBuilder $container)
...
-        $this->loadMetadata($config, $loader, $container);
+        $this->loadMetadata($loader, $container);
+        $this->loadPersistence($config['persistence'], $loader);
+    }
+
+    private function loadPersistence($config, Loader\XmlFileLoader $loader)
+    {
+        foreach ($config as $persistenceName => $config) {
+            if (false === $config['enabled']) {
+                return;
```
danrot: Shouldn't this be continue? Otherwise it will stop for the first configuration which is not enabled.

README.rst
```diff
@@ -103,16 +103,17 @@ Mapping
 -------
 
 The MassiveSearchBundle requires that you define which objects should be indexed
-through *mapping*. Currently only **XML mapping** supported:
+through *mapping*. Currently only **XML mapping** is natively supported:
```
danrot: What do you want to say by natively?

README.rst
```diff
@@ -103,16 +103,17 @@ Mapping
...
 
     <!-- /path/to/YourBundle/Resources/config/massive-search/Product.xml -->
     <massive-search-mapping xmlns="http://massive.io/schema/dic/massive-search-mapping">
 
         <mapping class="Massive\Bundle\SearchBundle\Tests\Resources\TestBundle\Entity\Product">
-            <indexName>product</indexName>
-            <idField name="id"/>
+            <index name="product" />
+            <id property="id" />
+            <locale property="locale" />
```
danrot: That's something that definitely should go into the UPGRADE.md

README.rst
```diff
@@ -122,10 +123,124 @@ through *mapping*. Currently only **XML mapping** supported:
     </massive-search-mapping>
 
 This mapping will cause the fields ``title`` and ``body`` to be indexed into
-an index named ``product`` using the ID obtained from the objects ``id``
-field. (We use the Symfony `PropertyAccess
-<http://symfony.com/doc/current/components/property_access/index.html>`_
-component, so it works on properties and methods alike).
+an index named ``product`` using the ``locale`` determined from the locale
+field and the ID obtained from the objects ``id``
+field.
```
danrot: Why goes field into a new line?

README.rst
dantleech: XML is supported natively (i.e. it comes with XML support) - but it implies that you can add your own metdata drivers.

README.rst
```diff
@@ -122,10 +123,124 @@ through *mapping*. Currently only **XML mapping** supported:
...
+            <index name="product" />
+            <id property="id" />
+            <url expr="'/path/to/' ~ article.title'" />
+            <fields>
+                <field name="title" type="string" />
+                <field name="body" type="string" />
+            </fields>
+
+            <context name="admin">
+                <url exp="'/admin/edit/article/' ~ object.id" />
```
danrot: I am still not a fan of this :-/ It couples the backend too much to the frontend... @3rfan Do you build now this URL building stuff we have talked about? Would make this somehow obsolete...

Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -68,8 +68,6 @@ public function __construct(
      */
     public function index(Document $document, $indexName)
     {
-        $indexName = $this->localizationStrategy->localizeIndexName($indexName, $document->getLocale());
```
danrot: Just curious: what has this been replaced with?

Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -183,14 +167,83 @@ public function search(SearchQuery $searchQuery)
...
 
+    /**
+     * {@inheritDoc}
+     */
     public function getStatus()
     {
+        $indexes = $this->listIndexes();
+
         $indices = $this->client->indices()->status(array('index' => '_all'));
+        $indexes = $indices['indices'];
```
danrot: is indexes a real word? :-)

Search/Adapter/ElasticSearchAdapter.php
dantleech: All the localization has been moved "up" to the SearchManager. The adapters are now localization agnostic.

Search/Adapter/ElasticSearchAdapter.php
dantleech: yep: http://en.wiktionary.org/wiki/indexes :)

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -26,16 +26,12 @@
...
  */
 class ZendLuceneAdapter implements AdapterInterface
 {
     const ID_FIELDNAME = '__id';
     const CLASS_TAG = '__class';
 
-    // TODO: This fields should be handled at a higher level
     const URL_FIELDNAME = '__url';
     const TITLE_FIELDNAME = '__title';
     const DESCRIPTION_FIELDNAME = '__description';
```
danrot: Couldn't these fields be at least defined in the `AdapterInterface`? Then you don't have to define them twice.

Search/Event/IndexRebuildEvent.php
```diff
@@ -0,0 +1,62 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search\Event;
+
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Output\NullOutput;
+
+class IndexRebuildEvent extends Event
```
danrot: PHPDoc missing

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
```diff
@@ -0,0 +1,155 @@
...
+use Doctrine\Common\Persistence\ObjectManager;
+use Massive\Bundle\SearchBundle\Search\Event\IndexRebuildEvent;
+use Metadata\MetadataFactory;
+use Massive\Bundle\SearchBundle\Search\SearchManager;
+use Symfony\Component\Console\Output\OutputInterface;
+use Doctrine\Common\Persistence\Mapping\ClassMetadata as OrmMetadata;
+use Massive\Bundle\SearchBundle\Search\Metadata\IndexMetadata;
+use Massive\Bundle\SearchBundle\Search\Metadata\ClassMetadata;
+
+class DoctrineOrmIndexRebuildSubscriber implements EventSubscriberInterface
```
danrot: PHPDoc missing

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
```diff
@@ -0,0 +1,155 @@
...
+    {
+        $output = $event->getOutput();
+        $filter = $event->getFilter();
+        $purge = $event->getPurge();
+
+        $metadataFactory = $this->objectManager->getMetadataFactory();
+        $metadata = $metadataFactory->getAllMetadata();
+
+        foreach ($metadata as $class) {
+            if ($filter && !preg_match('{' . $filter . '}', $class->name)) {
```
danrot: What are you filtering here for?

Command/IndexRebuildCommand.php
danrot: We don't do that anywhere in Sulu, because many people can merge PRs here or edit the file in other ways, so this information is quite useless, we have Git for that. For consistency I would prefer to remove these tags.

Search/EventSubscriber/DoctrineOrmSubscriber.php
```diff
@@ -0,0 +1,100 @@
...
+use Doctrine\Common\EventSubscriber;
+use Doctrine\ORM\Event\LifecycleEventArgs;
+use Doctrine\ORM\Events;
+use Massive\Bundle\SearchBundle\Search\SearchManagerInterface;
+use Massive\Bundle\SearchBundle\Search\Exception\MetadataNotFoundException;
+
+/**
+ * Index mapped doctrine ORM documents
+ *
+ * @author Daniel Leech <daniel@dantleech.com>
```
danrot: Same as above, I would remove these tags.

Search/Exception/MetadataNotFoundException.php
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search\Exception;
+
+class MetadataNotFoundException extends \InvalidArgumentException
```
danrot: PHPDoc missing, and I am not sure if the `InvalidArgumentException` is the correct parent here... And I would add more information to this exception (which metdata has not been found?)

Command/IndexRebuildCommand.php
dantleech: this isn't Sulu ;-) But I don't mind, I am just +1 for authorship tags in open source.

Search/ExpressionLanguage/MassiveSearchExpressionLanguage.php
```diff
@@ -0,0 +1,89 @@
...
+                return sprintf('join(%s, %s)', $glue, $elements);
+            },
+            function (array $values, $glue, $elements) {
+                return implode($glue, $elements);
+            }
+        );
+    }
+
+    /**
+     * Join is an analogue for array_map. The callback
```
danrot: I guess Join is not correct here

Search/ExpressionLanguage/MassiveSearchExpressionLanguage.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    private function createMapFunction()
+    {
+        return new ExpressionFunction(
+            'map',
+            function ($elements, $expression) {
+                throw new \Exception('Map function does not support compilation');
+            },
+            function (array $values, $elements, $expression) {
+                $expressionLanguage = $this;
```
danrot: What do you need this line for? Why not just using `$this`?

Search/Factory.php
```diff
@@ -49,13 +53,37 @@ public function makeField($name, $value, $type = Field::TYPE_STRING)
     }
 
     /**
-     * Make a new metadata object representing a mapped
-     * searchable class
+     * Make a new metadata object repesenting a mapped
+     * searchanle class
```
danrot: searchable

Search/Factory.php
```diff
@@ -49,13 +53,37 @@ public function makeField($name, $value, $type = Field::TYPE_STRING)
     }
 
     /**
-     * Make a new metadata object representing a mapped
-     * searchable class
+     * Make a new metadata object repesenting a mapped
```
danrot: representing

Search/Factory.php
```diff
@@ -49,13 +53,37 @@ public function makeField($name, $value, $type = Field::TYPE_STRING)
...
+    {
+        return new IndexMetadata();
+    }
+
+    public function makeMetadataField($name)
+    {
+        return new MetadataField($name);
+    }
+
+    public function makeMetadataProperty($path)
```
danrot: Is this file registered as a service? These methods seem a bit strange to me, why not just using some static methods on these classes?

Search/Metadata/Driver/XmlDriver.php
```diff
@@ -14,25 +14,42 @@
 use Metadata\Driver\AbstractFileDriver;
 use Massive\Bundle\SearchBundle\Search\Factory;
 use Metadata\Driver\FileLocatorInterface;
+use Massive\Bundle\SearchBundle\Search\Metadata\Field\Property;
+use Massive\Bundle\SearchBundle\Search\Metadata\Field\Expression;
+use Massive\Bundle\SearchBundle\Search\Metadata\Field\Field;
 
 class XmlDriver extends AbstractFileDriver implements DriverInterface
```
danrot: Missing PHPDoc

Search/ExpressionLanguage/MassiveSearchExpressionLanguage.php
dantleech: I think that should have been declareed outside of the clouse and then passed with `use`, as older versions of PHP do not support `$this` inside a closure.

Search/Metadata/Field/Field.php
```diff
@@ -0,0 +1,53 @@
...
+ * This is different from Property in that this represents
+ * a value that should be determined from the "name" of the
+ * mapped field instead of an explicit property declaration.
+ *
+ * For example:
+ *
+ *     <field name="foobar" /> <!-- yields a Field -->
+ *     <field name="foobar" property="foo" /> <!-- yields a Property -->
+ *
+ * @author Daniel Leech <daniel@dantleech.com>
```
danrot: Would omit this tag :-)

Search/Metadata/Field/Property.php
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Massive\Bundle\SearchBundle\Search\Metadata\Field;
+
+use Massive\Bundle\SearchBundle\Search\Metadata\FieldInterface;
+
+/**
+ * Simple value object for representing properties which
+ * should be accesible using the PropertyAccessor
+ *
+ * @author Daniel Leech <daniel@dantleech.com>
```
danrot: Would omit this tag :-)

Search/Metadata/FieldEvaluator.php
```diff
@@ -0,0 +1,125 @@
...
+    {
+        try {
+            switch (get_class($field)) {
+                case 'Massive\Bundle\SearchBundle\Search\Metadata\Field\Property':
+                    return $this->getPropertyValue($object, $field);
+                case 'Massive\Bundle\SearchBundle\Search\Metadata\Field\Expression':
+                    return $this->getExpressionValue($object, $field);
+                case 'Massive\Bundle\SearchBundle\Search\Metadata\Field\Field':
+                    return $this->getFieldValue($object, $field);
+            }
```
danrot: I would use lines like `Property::class`. Makes future refactorings easier :-)

Search/Metadata/FieldEvaluator.php
```diff
@@ -0,0 +1,125 @@
...
+                'object' => $object,
+            ));
+        } catch (\Exception $e) {
+            throw new \RuntimeException(sprintf(
+                'Error encountered when evaluating expression "%s"',
+                $field->getExpression()
+            ), null, $e);
+        }
+    }
+}
```
danrot: IMHO you use too many general exceptions in this class (Exception, RuntimeException)

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Populate the Document with the actual values from the object which
+     * is being indexed.
+     *
+     * @param  Document                  $document
+     * @param  mixed                     $object
+     * @param  array                     $fieldMapping
+     * @param  string                    $prefix       Prefix the document field name (used when called recursively)
+     * @throws \InvalidArgumentException
```
danrot: Please don't align the PHPDoc like this.

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,188 @@
...
+                }
+
+                $childObjects = $this->fieldEvaluator->getValue($object, new Property($fieldName));
+
+                foreach ($childObjects as $i => $childObject) {
+                    $this->populateDocument(
+                        $document,
+                        $childObject,
+                        $mapping['mapping']->getFieldMapping(),
+                        $prefix.$fieldName.$i
```
danrot: Codestyle: spaces between operators and operands.

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,188 @@
...
+
+                continue;
+            }
+
+            $value = $this->fieldEvaluator->getValue($object, $mapping['field']);
+
+            if (!is_array($value)) {
+                $document->addField(
+                    $this->factory->makeField(
+                        $prefix.$fieldName,
```
danrot: Codestyle: spaces

Search/ObjectToDocumentConverter.php
```diff
@@ -0,0 +1,188 @@
...
+                    )
+                );
+
+                continue;
+            }
+
+            foreach ($value as $key => $itemValue) {
+                $document->addField(
+                    $this->factory->makeField(
+                        $prefix.$fieldName.$key,
```
danrot: Codestyle: spaces

Controller/RestController.php
```diff
@@ -0,0 +1,56 @@
...
+use Massive\Bundle\SearchBundle\Search\SearchManagerInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * REST API for search
+ *
+ * @author Daniel Leech <daniel@dantleech.com>
+ */
+class RestController
```
danrot: I would not call this one `RestController`, `SearchController` seems like the better choice.

Command/IndexRebuildCommand.php
danrot: It's not Sulu, but I am totally against handling this different in every project :-P And I don't like it for the above reasons.

Search/ExpressionLanguage/MassiveSearchExpressionLanguage.php
danrot: The latest still maintained version (5.4) supports `$this` in closures, so you don't have to think about this ;-) Just use this in closures.

Search/Factory.php
dantleech: This is a service that enables the user to implement their own classes, for example, the Metadata.

It is not very common, but is required if you want flexibility instead of having classes instantiate their own objects, and it also protects BC - for example in the Sulu Metadata driver we need to instantiate new fields, if we did that manually it would mean that we could not move, rename or change the constructor arguments of the MassiveSearch classes.

danrot: Ok, that's fine for me. But I think it is uncommon to have one factory class for all of them... But the classes are connected in some way, so it seems to be ok. However, don't you think that `create` is a more common name than `make`?

Command/IndexRebuildCommand.php
dantleech: Sulu is an exception in the Symfony ecosystem in this respect :) But like I said, I don't mind

dantleech: This guy even puts his name in the header .. https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpFoundation/Request.php#L6

Search/Factory.php
dantleech: Create is more common in that I have seen that used before, I choose `make` after looking at [wikipedia](http://en.wikipedia.org/wiki/Factory_method_pattern). Its got less characters than create and factorys "make" things in the real world. Normally you wouldn't say that a factory "creates" things. So in balance, I prefer "make".

danrot: I just checked a few factory classes in Symfony, and they are using create. I would prefer to use this word, since it is more common, and what people expect.

dantleech: In anycase, this is out of the scope of this PR, this was introduced in master a while back.

Command/QueryCommand.php
dantleech: You can specify either a single `index` or multiple `indexes`. The query command now supports multiple indexes.

Controller/RestController.php
dantleech: Becuase its a big extra dependency. this is just a simple implementation. If people want more they can do it themselves.

DependencyInjection/MassiveSearchExtension.php
dantleech: Yep.

README.rst
dantleech: Its in the Sulu upgrade, but not this one. will add.

README.rst
dantleech: This is an example and not specific to Sulu. It is valid.

Search/Adapter/ZendLuceneAdapter.php
dantleech: They are adapter specific. For example an adapter might have a better way of namespacing properties than this. The adapters are under no obligation to use the same names, and I don't want to imply that they should.

Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
dantleech: It filters the class name.

dantleech: e.g. so that you can choose to only rebuild Contacts, f.e.

Search/Exception/MetadataNotFoundException.php
dantleech: The exception message is passed in the constructor. RuntimeException is better I think.

Search/ExpressionLanguage/MassiveSearchExpressionLanguage.php
dantleech: Added 5.4 as the minimum version in composer.

Search/Metadata/FieldEvaluator.php
dantleech: That would push the minimum php version up to 5.5. This would suck, for example my server only has 5.4

Search/Metadata/FieldEvaluator.php
dantleech: ExpressionLanguage could throw a number of Exceptions. I wrap any exception that it does throw. I think that is fine.

Search/ObjectToDocumentConverter.php
dantleech: Ah, it was the cs fixer.

Controller/RestController.php
dantleech: Changed.

dantleech:
Updated.

UPGRADE.md
```diff
@@ -1,9 +1,14 @@
 UPGRADE
 =======
 
+
+
 0.5
 ---
 
+- XML mappings have changed, the "Field" suffix has been dropped, for example
+  ``<idField ... />`` is now ``<id ... />``.
```
danrot: `<index/>` has also changed.

Search/Exception/MetadataNotFoundException.php
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search\Exception;
+
+class MetadataNotFoundException extends \RuntimeException
```
danrot: I would still also pass the class for which no metadata was found to the constructor, so that you can act accordingly to it in the catch block. Just an error message is too less IMHO

Search/Metadata/FieldEvaluator.php
danrot: But that makes catching exceptions quite hard and not really intuitive.

README.rst
dantleech: Added.

Search/Metadata/FieldEvaluator.php
dantleech: In what case would you need to handle such an exception? It means that your system is broken and you need to fix it. I would rather add more explict exception classes when they are required.

Search/Exception/MetadataNotFoundException.php
dantleech: Again I would rather do this if there is a requirement. I am not  a fan of overriding the constructor arguments of Exceptions in general. The exception message is the single most important part of an exception - hiding it in the exception itself often means that the exception ends up making no sense as people do not see it.

dantleech:
Updated.

Search/Metadata/FieldEvaluator.php
danrot: Looks like we've got different opinions on this ;-) Especially in such an abstract library it is important for me to throw typed exceptions.

Search/Exception/MetadataNotFoundException.php
danrot: I would also do it because of a different reason, then you can set the exception text in one single place, and it's always the same. If you throw this exception at different locations with different texts it can be very confusing for the developer, which uses this library.

dantleech: Well, if you are throwing the exception in different places with the same meaning then you have a [SPOT](http://en.wikipedia.org/wiki/Single_Point_of_Truth) problem, if you are throwing the same exception but with a different context, then the message should be different in each case.

dantleech:
Have removed the REST API, and added category support.

Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -79,7 +78,7 @@ public function index(Document $document, $indexName)
                     break;
                 default:
                     throw new \InvalidArgumentException(sprintf(
-                        'Search field type "%s" is not know. Known types nare: %s',
+                        'Search field type "%s" is not know. Known types are: %s',
```
aitboudad: sprintf require here two arguments:
```php
throw new \InvalidArgumentException(sprintf(
    'Search field type "%s" is not know. Known types are: %s',
    $field->getType(),
    implode(', ', Field::getValidTypes())
));
```

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -222,24 +212,51 @@ public function getStatus()
...
+    }
 
-        return $this->getLuceneIndex($indexName);
+    /**
+     * {@inheritDoc}
+     */
+    public function listIndexes()
+    {
+        $finder = new Finder();
+        $indexDirs = $finder->directories()->depth('== 0')->in($this->basePath);
```
aitboudad: check if basePath exist before doing this

Command/QueryCommand.php
```diff
@@ -66,10 +64,10 @@ public function execute(InputInterface $input, OutputInterface $output)
                 $this->truncate($document->getDescription(), 50),
                 $document->getUrl(),
                 $document->getImageUrl(),
-                $document->getClass()
+                $document->getClass(),
             ));
         }
-        $table->render($output);
+        $table->render();
```
aitboudad: when there is no results, I prefer a simple msg like There are no results available for the given query" instead of rendering an empty table.

Search/SearchManager.php
```diff
@@ -178,132 +190,229 @@ public function getStatus()
...
+     * for the given index name, optionally limiting to the given locale.
+     *
+     * @param string $indexName
+     * @param string $locale
+     *
+     * @return string[]
+     */
+    private function getLocalizedIndexNamesFor($indexName, $locale = null)
+    {
+        $adapterIndexNames = $this->adapter->listIndexes();
```
aitboudad: if we use listIndexes from adapter we need to exlude indexes who doesn't exist in the metadata

Search/Adapter/ElasticSearchAdapter.php
dantleech: Indeed, should add a test for that.

Search/Adapter/ZendLuceneAdapter.php
dantleech: It now will return an empty array if the directory does not exist.

Command/QueryCommand.php
dantleech: Added a count of results and the time elapsed instead

````
+-------+----+-------+-------------+-----+-------+-------+
| Score | ID | Title | Description | Url | Image | Class |
+-------+----+-------+-------------+-----+-------+-------+
0 result(s) in 0.161362s
````

Make sense?

Search/SearchManager.php
dantleech: This private method is passed the `$indexName`, which at this point has either been submitted by the user (and validated see `validateQuery`) or was already determined from the metadata.

dantleech:
Thanks @aitboudad I have added tests for invalid field mapping.

Command/QueryCommand.php
aitboudad: yes, why not :)

Search/SearchManager.php
aitboudad: if I remove an index from metadata but the data still exist with Zend Lucene listIndexes return the deleted index !!

dantleech: It does, but you will get an exception if you try and search for it. We should probably add a "clean" command to remove unused indexes.

dantleech: https://github.com/massiveart/MassiveSearchBundle/issues/25

aitboudad: to get an exception you must call validateQuery after expandQueryIndexes in https://github.com/massiveart/MassiveSearchBundle/blob/feature/rest-api/Search/SearchManager.php#L149-L150

dantleech: The could supply the index name in the query, f.e. `products`. Expand indexes will add additional localized indexes if they exist, e.g. `products_de`, `products_fr`. These expanded names are not referenced in the metadata. If you search for an index not referenced in the metadata, you should get an exception.

aitboudad: ahh it's ok now, thanks :+1:

Search/Exception/MetadataNotFoundException.php
danrot: Haven't heard about SPOT before, but from what I have read there it is not an issue here, since the exception is only for the developer. I would definitely add more information here, but it's not that important to me, since we can add it also later.

danrot:
Will this still work with the old search stuff in Sulu? Or should it be merged together with the Sulu PR?

Search/Exception/MetadataNotFoundException.php
dantleech: All the information is in the place where the exception is thrown.

dantleech:
@danrot people should be using the tagged version so it can be merged at any time.

Search/Exception/MetadataNotFoundException.php
danrot: That's not where it is useful ;)

dantleech: Why is it useful here?

danrot: I said it's not useful here. But the information which metadata was not found might be useful in the place the exception is caught. And if it is, I don't want to use some string operations on the exception message to find out.

danrot:
Merged, but Sulu won't work with this. For the next Sulu release we have to tag 0.5.0 and https://github.com/sulu-io/sulu/pull/797 must be merged.

PR #19, opened 2015-02-06
Fixed elastic DI config

dantleech:
The current Elastic DI configuration is broken, added a test.


Tests/Unit/DependencyInjection/MassiveSearchExtensionTest.php
```diff
@@ -131,5 +132,8 @@ public function testAdapterConfig($adapterId, $config, $expectedParameters)
                 $this->container->getParameter($expectedKey)
             );
         }
+
+        $serviceId = 'massive_search.adapter.' . $adapter;
+        $this->container->get($serviceId);
```
dantleech: Here explicitly check that the adapter is instantiable.

PR #24, opened 2015-03-14
[Search field] remove unsupported type.

aitboudad:


dantleech:
Thanks. I think this should actually be removed for the time being as the metadata driver currently only maps string values.

aitboudad:
Ok, I removed TYPE_BINRARY.

PR #27, opened 2015-04-23
Allow configuration of default zend lucene character encoidng

dantleech:
Allows proper indexing and searching of international characters with Zend Lucene.


dantleech:
Updated.

PR #31, opened 2015-05-06
Order by score when searching with the Zend adapter

dantleech:
The Zend Lucene `MultiSearcher` does not currently support ordering. This PR does a `usort` over the query hits to order them by score.


ampaze:
Am I missing something or is this really sorting the highest score to the end?

alexander-schranz:
This implementation was correct but it was accidently changed to a spaceship operator here: https://github.com/massiveart/MassiveSearchBundle/pull/163 which did make it fail. effected versions are 2.6.3 and 2.6.4

PR #32, opened 2015-05-07
Added index strategy

dantleech:
This PR introduces the possilbity to define an INDEX_STRATEGY.

This determines how user fields are stored in the search implementation, currently:

- `INDEX_AGGREGATE`: Index the fields in a single aggregate field, but store them individually (for performance)
- `INDEX_UNSTORED`: Index the field but do not store it.

This is currently only applying to Zend Lucene. Elastic Search will index all fields separtely.

Note these things are currenlty only exposed to the Drivers. No support is avaiilable to map these fields.

### DIFF
--- a/Search/Field.php
+++ b/Search/Field.php
@@ -30,7 +30,26 @@ class Field
      */
     protected $value;
 
+    /**
+     * @var string
+     */
+    protected $indexStrategy;
+
+    /**
+     * Store the field as a string
+     */
     const TYPE_STRING = 'string';
+
+    /**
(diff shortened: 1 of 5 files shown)


Search/Field.php
```diff
@@ -30,8 +30,27 @@ class Field
...
     const TYPE_STRING = 'string';
 
+    /**
+     * Aggregate the fields in a single aggregate field for indexing and
+     * store the fields
+     */
+    const INDEX_AGGREGATE = 'aggregate';
+
+    /**
+     * Index, but do not sture the field
```
danrot: store

Issue #38, opened 2015-05-08
Searching for non-existing indexes can pass an empty array to an adapter and cause a global search


dantleech:
Currently localized index names are inferred from existing indexes, however if the indexes do not exist then no indexes are inferreed.

Adapters tend to perform global searches when no indexes are given.

This needs to be fixed.

dantleech:
Done.

PR #39, opened 2015-05-08
Temporarily disabled global search feature of test adapter

dantleech:
Disabling this feature for now as it is not required by any tests and it causes a regression in the Sulu tests.

See https://github.com/massiveart/MassiveSearchBundle/issues/38


dantleech:
Note that infact this is valid behavior -- the adapter should never be passed an empty set of indexes.

PR #40, opened 2015-05-08
Allow indexed and stored field mapping for Zend Lucene

dantleech:
Adds the posibility to explicitly map a field to be both indexed and stored.

Note this currently only applies to ZendLucene.


dantleech:
See https://github.com/massiveart/MassiveSearchBundle/issues/41 for a better way to implement indexing policy.

wachterjohannes:
@dantleech i will make a hotfix release with this tomorrow!

Issue #41, opened 2015-05-08
Explicitly use `index` and `store` booleans instead of the composite types inherited from ZendLucene.


dantleech:
Instead of `unStored`, `stored`, `text`, etc. Use `store` and `index`.

Appplies to the `Index::INDEX_*` constants.

dantleech:
Done

PR #44, opened 2015-05-11
Use "store" and "index" fields to determine how fields are indexed

dantleech:
Fixes: https://github.com/massiveart/MassiveSearchBundle/issues/41

Use boolean fields `store` and `indexed` and `aggregate` to replace the single, non-web-scale, `index strategy` field.

### DIFF
--- a/Search/Adapter/ZendLuceneAdapter.php
+++ b/Search/Adapter/ZendLuceneAdapter.php
@@ -63,30 +63,22 @@ class ZendLuceneAdapter implements AdapterInterface
     private $encoding;
 
     /**
-     * @var string
-     */
-    private $defaultIndexStrategy;
-
-    /**
      * @param Factory $factory
      * @param string $basePath Base filesystem path for the index
      * @param bool $hideIndexException
      * @param null $encoding
-     * @param string $defaultIndexStrategy
      */
     public function __construct(
         Factory $factory,
         $basePath,
         $hideIndexException = false,
-        $encoding = null,
-        $defaultIndexStrategy = Field::INDEX_AGGREGATE
+        $encoding = null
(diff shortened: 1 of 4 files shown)


Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -393,18 +379,31 @@ private function getIndex($indexPath, $create = false)
...
+
+        if (false === $field->isStored() && $field->isIndexed()) {
+            return 'unStored';
+        }
+
+        if ($field->isStored() && false === $field->isIndexed()) {
+            return 'unIndexed';
+        }
+
+        if (false === $field->isStored() && false === $field->isIndexed()) {
```
dantleech: It couldn't be anything else ?

dantleech:
I think this makes sense as elastic also has `index` and `store` fields which correspond to the options we are adding here.

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -330,6 +300,22 @@ public function flush(array $indexNames)
...
     /**
+     * Optimize the search indexes after the index rebuild event has been fired.
+     * Should have a priority low enough in order for it to be executed after all
+     * the actual index builders.
+     *
+     * @param IndexRebuildEvent $event
+     */
+    public function optimizeIndexAfterRebuild(IndexRebuildEvent $event)
+    {
+        foreach ($this->listIndexes() as $indexName) {
```
dantleech: This is not part of this PR?

dantleech:
Note that this is a BC break for Sulu, but not a big change. I suggest we merge this now. /cc @chirimoya

PR #53, opened 2015-05-26
Added benchmarks

dantleech:
Compares:
- Zend Lucene
- Elastic Search
- Test Adapter (in memory adapter)


dantleech:
The options-resolver dependency is too low on PHPBench.

PR #60, opened 2015-10-08
Added support for arrays

dantleech:
This PR adds support for indexing arrays.

ElasticSearch supports this natively (no code change required), With Zend we emulate it with a space-delimited list. This does however mean that spaces are not permitted in the array values - one way to solve this might be to make a custom analyzer.


dantleech:
Cannot make a custom analyzer, but on reflection I think perhaps this is not an issue. This is a search engine, and `array("Jack Jackson", "Jill")` should tokenize both "Jack" and "Jackson" and "Jill".

CHANGELOG.md
```diff
@@ -1,6 +1,14 @@
 CHANGELOG
 =========
 
+dev-develop
+-----------
+
+### Feature
```
danrot: Can you maybe convert the entire file to use the markdown syntax, as I did in Husky? It's somehow embarrassing that all our repos have different styleguides on this :confused:

Resources/docs/extending.rst
```diff
@@ -78,6 +78,23 @@ definition.
     Adding new metadata providers is non-trivial, you should check the
     existing code for implementation details.
 
+Converters
+----------
+
+Massive Search allows you to implement custom types. This types can be defined
+with custom converters which converts the value of a field into a indexable
```
danrot: `custom converters, which`

Resources/docs/mapping.rst
```diff
@@ -62,6 +62,10 @@ Types:
 - ``string``: Store as a string
 - ``complex``: Apply mapping to an array of values
 
+Additionally you can define your own types with custom converters (see
+:doc:`extending`). This converters converts the value into a indexable
```
danrot: `These converters convert the value into an indexable`

Search/Converter/ConverterManagerInterface.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Massive\Bundle\SearchBundle\Search\Converter;
+
+/**
+ * Interface for converter manager.
+ */
+interface ConverterManagerInterface
+{
+    /**
+     * Converts value from a source format to a target format.s.
```
danrot: `formats`

Search/Converter/ConverterManager.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->converter[$from] = $converter;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function convert($value, $from)
+    {
+        if (!$this->hasConverter($from)) {
+            throw new NoConverterFoundException($from);
```
danrot: I would call this exception `ConverterNotFoundException` according to the naming of other exceptions, wouldn't you?

Search/Converter/ConverterManagerInterface.php
```diff
@@ -0,0 +1,36 @@
...
+    public function convert($value, $from);
+
+    /**
+     * Returns true if a converter exists which converts from source to target format.
+     *
+     * @param string $from source format.
+     *
+     * @return bool
+     */
+    public function hasConverter($from);
```
danrot: Where's the target here? It should be here according to the comment above. Same question for the other method.

Search/Event/DeindexEvent.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    private $subject;
+
+    public function __construct($subject)
+    {
+        $this->subject = $subject;
+    }
+
+    /**
+     * @return object
```
danrot: I think the comment here would be more important than on the variable, since this is the method the user uses.

Search/Event/IndexEvent.php
```diff
@@ -0,0 +1,39 @@
...
+     */
+    private $subject;
+
+    public function __construct($subject)
+    {
+        $this->subject = $subject;
+    }
+
+    /**
+     * @return object
```
danrot: Same as in previous file.

Search/ObjectToDocumentConverter.php
```diff
@@ -170,21 +182,31 @@ private function populateDocument($document, $object, $fieldMapping, $prefix = '
...
+                $value = $this->converterManager->convert($value, $type);
+
+                $type = is_array($value) ? Field::TYPE_ARRAY : Field::TYPE_STRING;
             }
 
-            if (!is_array($value)) {
+            if ($value !== null && false === is_scalar($value) && false === is_array($value)) {
+                throw new \InvalidArgumentException(
+                    sprintf(
+                        'Search field "%s" resolved to non-scalar value with type "%s". Only scalar (single) values can be indexed.',
```
danrot: The error message is wrong. And are there still any fields not indexable? Is it possible at all that an object is passed into here?

Search/Converter/ConverterManagerInterface.php
wachterjohannes: i have removed the target ... the target is an indexable format (string or array) i will adopt the comment.

PR #61, opened 2015-10-13
Added reindex directive

dantleech:
This PR introduces a way to specify which method to use on the entity repository when retrieving entities which should be reindexed.

This provides a way to only reindex certain objects for certain class mappings.

````xml
<?xml version="1.0" ?>
<mapping ...>
    <reindex repository-method="findOnlySomeStuff" />
</mapping>
````

NOTE: This PR only introduces the `repository-method` mapping, but it follows that there can be introduced a `repository-id` mapping which specifies a service which could be used to instead of the, in this case, default ORM repository.

### DIFF
--- a/Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
+++ b/Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
@@ -84,14 +84,14 @@ class DoctrineOrmIndexRebuildSubscriber implements EventSubscriberInterface
         $purge = $event->getPurge();
 
         $metadataFactory = $this->objectManager->getMetadataFactory();
-        $metadata = $metadataFactory->getAllMetadata();
+        $metadatas = $metadataFactory->getAllMetadata();
 
-        foreach ($metadata as $class) {
+        foreach ($metadatas as $class) {
             if ($filter && !preg_match('{' . $filter . '}', $class->name)) {
                 continue;
             }
 
-            $searchMeta = $this->searchMetadataFactory->getMetadataForClass($class->getName());
+            $searchMeta = $this->searchMetadataFactory->getMetadataForClass($class->name);
 
             if (null === $searchMeta) {
                 continue;
@@ -139,11 +139,21 @@ class DoctrineOrmIndexRebuildSubscriber implements EventSubscriberInterface
      */
     private function rebuildClass(OutputInterface $output, ClassMetadata $class)
     {
-        $output->write('<comment>Rebuilding</comment>: ' . $class->getName());
+        $output->write('<comment>Rebuilding</comment>: ' . $class->name);
 
         $repositoryMethod = $class->getReindexRepositoryMethod();
         $repositoryMethod = $repositoryMethod ?: 'findAll';
-        $objects = $this->objectManager->getRepository($class->name)->$repositoryMethod();
+
+        $repository = $this->objectManager->getRepository($class->name);
+
+        if (!method_exists($repository, $repositoryMethod)) {
+            throw new \InvalidArgumentException(sprintf(
+                'Repository method "%s" does not exist.',
+                $repositoryMethod
+            ));
+        }
+
+        $objects = $repository->$repositoryMethod();
 
         $count = 0;
         foreach ($objects as $object) {
--- a/Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
+++ b/Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php
@@ -100,10 +100,10 @@ class DoctrineOrmIndexRebuildSubscriber implements EventSubscriberInterface
             $classMetadata = $searchMeta->getOutsideClassMetadata();
 
             if ($purge) {
-                $this->doPurge($output, $classMetadata, $class);
+                $this->doPurge($output, $classMetadata);
             }
 
-            $this->rebuildClass($output, $class);
+            $this->rebuildClass($output, $classMetadata);
         }
     }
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriber.php [hunk 1]
```diff
@@ -138,11 +137,13 @@ private function doPurge(OutputInterface $output, ClassMetadata $classMetadata,
...
-    private function rebuildClass(OutputInterface $output, OrmMetadata $ormMetadata)
+    private function rebuildClass(OutputInterface $output, ClassMetadata $class)
     {
-        $output->write('<comment>Rebuilding</comment>: ' . $ormMetadata->getName());
+        $output->write('<comment>Rebuilding</comment>: ' . $class->getName());
 
-        $objects = $this->objectManager->getRepository($ormMetadata->name)->findAll();
+        $repositoryMethod = $class->getReindexRepositoryMethod();
+        $repositoryMethod = $repositoryMethod ?: 'findAll';
+        $objects = $this->objectManager->getRepository($class->name)->$repositoryMethod();
```
dantleech: Add check to see if method exists.

dantleech:
We seem to be missing functional tests for reindexing

dantleech:
Added functional test for reindex command, added unit test for Doctrine subscriber, added docs.

dantleech:
CS fixes seem to break the build with PHP5.4.

Tests/Functional/Command/IndexRebuildCommandTest.php
```diff
@@ -0,0 +1,40 @@
...
+        $this->tester = new CommandTester($command);
+    }
+
+    /**
+     * This command just fires an event.
+     */
+    public function testCommand()
+    {
+        $this->tester->execute([
+        ]);
```
danrot: Nothing else to test here? If not i wouldn't break this into two lines.

Tests/Unit/Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriberTest.php
```diff
@@ -0,0 +1,157 @@
...
+        ));
+        $this->ormMetadata->name = 'stdClass';
+        $this->searchMetadata->name = 'stdClass';
+        $this->searchMetadataFactory->getMetadataForClass('stdClass')->willReturn($this->searchHierarchy->reveal());
+
+        $this->objectManager->getRepository('stdClass')->willReturn(new TestRepository());
+    }
+}
+
+class TestRepository
```
danrot: Why are you doing this? Why don't you use another mock for that?

danrot:
The tests are not failing because of the CS fixes, the `::class` keyword has been introduced in PHP 5.5. For me it would be fine to declare 5.5, especially since 5.4 won't even see security releases anymore. But we should clarify with @chirimoya.

chirimoya:
http://docs.sulu.io/en/latest/book/getting-started/system-requirements.html ... Sulu already require PHP 5.5. So is fine for me.

Tests/Unit/Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriberTest.php
dantleech: It needs concrete methods to test the Exception if the method does not
exist. Not sure if prophecy does that (tbh I didn't look).

danrot: I would try, from my experience until now Prophecy is quite good with this kind of stuff.

dantleech:
Hmm, that could be it. I would rather just expand the namespaces for
now and preserve BC.

danrot:
As said, I would switch to PHP 5.5, since there is no security support anymore: http://php.net/supported-versions.php

I don't want to be part of a project which worsens PHPs reputation more than necessary :wink:

dantleech:
Using the `::class` feature breaks BC and deviates from the standard currently employed in this PR. If we remove `::class` it should technically be done in a major release and it should be done everywhere.

Tests/Unit/Search/EventSubscriber/DoctrineOrmIndexRebuildSubscriberTest.php
dantleech: Could not find a way to do this with prophecy.

danrot:
According to semantic versioning you are allowed to BC breaks like this prior to 1.0.0

danrot:
I have tried to use that in https://github.com/sulu-io/sulu/pull/1711, but it doesn't solve our issue https://github.com/sulu-io/sulu/issues/1323

When a new file is uploaded, it will still double the search results...

danrot:
TBH, I am not sure anymore if this is the best idea... Again, it is quite easy to get different representations after reindexing, although I would expect it to stay almost the same (expect there was an error in the configuration, and the content of the index will change because of this).

I favor my idea of indexing the `File` and let the MassiveSearchBundle somehow know which languages it has to index more and more...

danrot:
Ah, didn't think about the id expr you told me... That's why we should have these discsussions on github issues :smile: 

For the future the `File` is still what should be index IMHO.

dantleech:
So everything is OK otherwise? Yes, we can still "change everything" before 1.0, you are correct. But I would still do that in a separate PR which changes this globally.

danrot:
Yeah, but I would like that to be merged together with https://github.com/sulu-io/sulu/issues/1323, to make sure that everything works as expected. I hope @chirimoya finds some time to do that...

PR #65, opened 2015-10-27
made index field evaluatable

danrot:
__tasks:__

- [x] test coverage (index field in search results)
- [x] gather feedback for my changes
- [x] submit changes to the documentation
- [x] add index name to search results in elastic search (check if obtainable from elastic result, or if it has to be saved as a field as in Zend Lucene)

__informations:__

| q                | a
| ---------------- | ---
| Fixed tickets    | none
| BC breaks        | index tag changed in xml mapping
| Documentation PR | none


wachterjohannes:
@danrot this don`t work for snippets (maybe phpcr does it) and medias ...

.travis.yml
```diff
@@ -6,7 +6,7 @@ services:
 sudo: false
 
 php:
-  - 5.4
+  - 5.5
```
wachterjohannes: shouldnt we add here also 5.6?

Resources/docs/introduction.rst
```diff
@@ -103,7 +103,7 @@ Place the following mapping file in the
     <massive-search-mapping xmlns="http://massive.io/schema/dic/massive-search-mapping">
 
         <mapping class="Model\Product">
-            <index name="product" />
+            <index value="product" />
```
wachterjohannes: why is this here now value - i think "name" would be a better name for this property!

Search/Document.php
```diff
@@ -260,7 +260,6 @@ public function jsonSerialize()
             'url' => $this->url,
             'image_url' => $this->imageUrl,
             'locale' => $this->locale,
-            'category' => $this->category,
```
wachterjohannes: you have replaced the category with index - didnt you need the index here?

Search/LocalizationStrategyInterface.php
```diff
@@ -20,18 +20,31 @@
      * Provide localized version of the given index name.
      *
      * @param string $indexName
+     * @param string $locale
+     *
+     * @return
```
wachterjohannes: empty return type

Search/LocalizationStrategyInterface.php
```diff
@@ -20,18 +20,31 @@
...
      */
     public function localizeIndexName($indexName, $locale);
 
     /**
+     * Remove the localization part of the given index name.
+     *
+     * @param $indexName
+     *
+     * @return mixed
```
wachterjohannes: i think this should be string?

.travis.yml
danrot: That's probably a good idea, will do it.

Resources/docs/introduction.rst
danrot: Because the `FieldEvaluator` evaluates the property differently based on the attribute name. I only adjusted the index tag to the other ones, where name is handled differently than on the index tag.

Search/Document.php
danrot: No, because the search document already knows from which index it is delivered.

PR #66, opened 2015-11-18
Added type 'date' to elastic search and zend lucene

stemod:


Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -78,6 +78,11 @@ public function index(Document $document, $indexName)
                 case Field::TYPE_ARRAY:
                     $fields[$massiveField->getName()] = $value;
                     break;
+                case Field::TYPE_DATE:
```
wachterjohannes: exactly for that i have introduced the converter mechanism why don`t you use it?

wachterjohannes: https://github.com/massiveart/MassiveSearchBundle/blob/develop/Resources/docs/extending.rst#converters

@danrot could it be that the docs wont be updated automatically?

danrot: Probably the docs are only rebuild for master. And if I think about it, we should maybe always use converter. So there could be a default converter for string and array, then we could skip this part of the code. Or am I overlooking something? Anyway, this wouldn't be part of this PR, but if we decide to do so, the date could be implemented as converter already.

wachterjohannes: this code snippet is only a check for provided types. for string or array i wouldnt create a converter. the string converter would return a string and the array converter would have to decide if he returns a array or a string (for lucene) ... therefor i would refactor this check here and not implement converter for this two types!

Search/Converter/Types/DateConverter.php
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search\Converter\Types;
+
+use Massive\Bundle\SearchBundle\Search\Converter\ConverterInterface;
+
+class DateConverter implements ConverterInterface
```
wachterjohannes: phpdoc

Search/Converter/Types/DateConverter.php
```diff
@@ -0,0 +1,24 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function convert($value)
+    {
+        return is_null($value) ? '' : $value->format('Y-m-d\TH:i:sP');
```
wachterjohannes: extract the format into a parameter! then it is possible to change it if someone wants that!

danrot: The format depends on the used adapter, so I wouldn't make that configureable. Although it makes sense to extract the value for sure (there are already some constants for date formats included, maybe we can even use these).

danrot:
@stemod @wachterjohannes The changelog was missing :wink:

wachterjohannes:
@stemod i will do it for you (:

PR #79, opened 2016-04-06
[FEATURE] Consider limit + offset for searching

### LINKED ISSUE #26: Support offset and max results in query
Introduce support for offest and max results in order to support pagination.

thomasduenser:
| Q | A
| --- | ---
| Bug fix? | no
| New feature? | yes
| BC breaks? | yes
| Deprecations? | no
| Fixed tickets | fixes #26
| Related issues/PRs | ---
| License | MIT

#### What's in this PR?

This Pull Request adds support for limit and offset. Additionally the returned value by the search contains the total results which is useful to build a pagination.

#### Why?

Because it was not possible to add a limit or offset to the search. Needed to build a pagination.

#### Example Usage

````
$hits = $this->getSearchManager()
            ->createSearch($queryString)
            ->locale('de')
            ->index('test')
            ->setOffset(10)
            ->setLimit(200)
            ->execute();
```

#### BC Breaks/Deprecations

The returned value was an array with the hits, now it's the class `QueryHits` which extends the `ArrayIterator` threw this change, the check `is_array()` doesn't work anymore.

#### To Do

- [x] Add Tests
- [x] Zend lucene adapter


Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -145,6 +146,8 @@ public function search(SearchQuery $searchQuery)
                     'query' => $queryString,
                 ],
             ],
+            'from' => $searchQuery->getOffset(),
```
wachterjohannes: how behaves this when offset and limit is the standard (i think `offset = 0; limit = null`)

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -203,9 +204,10 @@ public function search(SearchQuery $searchQuery)
 
         $hits = [];
 
-        foreach ($luceneHits as $luceneHit) {
+        for ($pos = $searchQuery->getOffset(); $pos < $searchQuery->getLimit(); $pos++) {
```
wachterjohannes: @dantleech do you know a better solution for that?

Search/QueryHits.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Search;
+
+class QueryHits extends \ArrayIterator
```
wachterjohannes: call this `SearchResult`

Search/Adapter/ZendLuceneAdapter.php
wachterjohannes: this is also wrong because if you had offset 100 and limit of 50 then you get nothing here. you should add here `min(count($luceneHits), $searchQuery->getOffset()+$searchQuery->getLimit())` to be sure that you don´t get `OutOfRangeException`

PR #80, opened 2016-04-07
added composer script handler for creating zend lucene directory

danrot:
This PR adds a new command `massive:search:init` and a composer script handler calling this command. This is useful to make the installation process easier, because the creation of this folder is already done by composer and does not have to be done manually. Another advantage is that always the correct folder based on the configuration is created.

### DIFF
--- a/Command/InitCommand.php
+++ b/Command/InitCommand.php
@@ -1,0 +1,46 @@
+<?php
+
+/*
+ * This file is part of the MassiveSearchBundle
+ *
+ * (c) MASSIVE ART WebServices GmbH
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Massive\Bundle\SearchBundle\Command;
+
+use Massive\Bundle\SearchBundle\Search\AdapterInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * Initializes the search adapter.
+ */
+class InitCommand extends Command
+{
+    /**
+     * @var AdapterInterface
+     */
+    private $adapter;
+
+    public function __construct(AdapterInterface $adapter)
+    {
+        parent::__construct();
+        $this->adapter = $adapter;
+    }
+
+    public function configure()
+    {
+        $this->setName('massive:search:init');
+        $this->setDescription('Initializes the search bundle');
+        $this->setHelp('This command will simply call the initialize method of the currently active search adapter.');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->adapter->initialize();
+    }
+}
--- a/Composer/SearchScriptHandler.php
+++ b/Composer/SearchScriptHandler.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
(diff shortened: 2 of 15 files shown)


dantleech:
Why a composer script? Surely it would be better to do this in the ZendLucene adapter?

danrot:
Because it should be automatically created after installation, similiar to the uploads folder in Sulu. And is there any other way than using a command here? Would you just add an `initialize` method to the `AdapterInterface` would probably make more sense, once I think about it again.

Command/InitCommand.php
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Creates the folder for the zend lucene search index if it does not exist yet.
+ */
+class InitCommand extends ContainerAwareCommand
+{
+    public function configure()
+    {
+        $this->setName('massive:search:init');
+        $this->setDescription('Initializes the data directory for the search bundle');
```
dantleech: Add help as with other commands.

```
$this->setHelp(<<<'EOT
This command does stuff

    $ %command.full_name&% arg1 arg2
EOT
);
```

dantleech: Add a functional test, it need only execute the command see `Tests\Functional\Command\` tests.

danrot: What help should I provide here? The command is not doing a lot, what it is doing basically fits into  the description.

Command/InitCommand.php
```diff
@@ -0,0 +1,45 @@
...
+    public function __construct(AdapterInterface $adapter)
+    {
+        parent::__construct();
+        $this->adapter = $adapter;
+    }
+
+    public function configure()
+    {
+        $this->setName('massive:search:init');
+        $this->setDescription('Initializes the search bundle');
```
dantleech: Add help

Tests/Unit/Search/Adapter/ZendLuceneAdapterTest.php
```diff
@@ -197,11 +203,32 @@ public function testIndexWithNullValues()
...
+    {
+        $adapter = $this->createAdapter($this->dataPath);
+
+        $this->filesystem->exists($this->dataPath)->willReturn(false);
+        $this->filesystem->mkdir($this->dataPath)->shouldBeCalled();
+
+        $adapter->initialize();
+    }
+
+    public function testInitializeWithExistingFolder()
```
dantleech: Would prefer if you added comments to these tests as with the others in the form of `It should ...`

dantleech:
Looks good, just a couple of small points.

danrot:
@dantleech Done.

Composer/SearchScriptHandler.php
```diff
@@ -0,0 +1,33 @@
...
+
+class SearchScriptHandler extends ScriptHandler
+{
+    /**
+     * @param $event CommandEvent A instance
+     */
+    public static function initBundle(CommandEvent $event)
+    {
+        $options = parent::getOptions($event);
+        $consoleDir = isset($options['symfony-bin-dir']) ? $options['symfony-bin-dir'] : $options['symfony-app-dir'];
```
dantleech: What would happen if neither of those options exist? Could that happen?

dantleech: The problem is that this would only work with a symfony-standard based project. 

We also have an implicit dependency on the `DistributionBundle` here which is bad.

PR #83, opened 2016-04-14
Updated dependencies for symfony 3

### LINKED ISSUE #81: Symfony 3 Support
I just tried to install your promising package on a symfony 3.0.4 installation and sadly failed because of the dependency on "symfony/property-access ~2.4". Are there any plans to support sf3 in the near future?

Thanks for your awesome work!

wachterjohannes:
This PR updates the dependencies to allow symfony 3.

Fixes #81

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,19 +5,26 @@ services:
 
 sudo: false
 
-php:
-  - 5.5
-  - 5.6
-  - 7.0
+matrix:
+  include:
+    - php: 5.5
+      env: PREFER="--prefer-lowest"
+    - php: 5.5
+      env: PREFER=""
+    - php: 5.6
+      env: PREFER=""
+    - php: 7.0
+      env: PREFER=""
 
 before_script:
     - composer selfupdate
-    - composer install
+    - composer update $PREFER
 
 script: 
     - vendor/bin/phpunit --coverage-text
     - vendor/bin/behat --suite=zend_lucene
     - vendor/bin/behat --suite=elastic
 
-env: 
-  - SYMFONY_DEPRECATIONS_HELPER=weak
+cache:
+  directories:
+    - $HOME/.composer/cache
--- a/behat.yml
+++ b/behat.yml
@@ -7,7 +7,7 @@ default:
                     adapterId: massive_search.adapter.zend_lucene
             bundle: MassiveSearchBundle
             filters:
-                tags: @zend_lucene
+                tags: '@zend_lucene'
         elastic:
             type: symfony_bundle
             contexts:
@@ -15,7 +15,7 @@ default:
                     adapterId: massive_search.adapter.elastic
             bundle: MassiveSearchBundle
             filters:
-                tags: @elastic
+                tags: '@elastic'
         test:
             type: symfony_bundle
             contexts:
@@ -23,7 +23,7 @@ default:
                     adapterId: massive_search.adapter.test
             bundle: MassiveSearchBundle
             filters:
-                tags: @test
+                tags: '@test'
     extensions:
         Behat\Symfony2Extension:
             kernel:
--- a/composer.json
+++ b/composer.json
@@ -10,20 +10,22 @@
     ],
     "require": {
         "jms/metadata": "~1.5",
-        "symfony/property-access": "~2.4",
+        "symfony/property-access": "~2.7||~3.0",
         "php": ">=5.4.0"
     },
     "require-dev": {
+        "phpunit/phpunit": "^4.5",
+
         "zendframework/zend-stdlib": "2.3.1 as 2.0.0rc5",
         "zendframework/zendsearch": "2.*@dev",
         "elasticsearch/elasticsearch": "~1.3",
         "symfony-cmf/testing": "dev-master",
-        "symfony/symfony": "~2.6",
-        "matthiasnoback/symfony-dependency-injection-test": "0.7.1",
-        "doctrine/doctrine-bundle": "~1.3",
-        "doctrine/orm": "~2.4",
+        "symfony/symfony": "~2.7||~3.0",
+        "matthiasnoback/symfony-dependency-injection-test": "0.7.*",
+        "doctrine/doctrine-bundle": "~1.4",
+        "doctrine/orm": "~2.5",
 
-        "behat/behat": "~3.0.0",
+        "behat/behat": "~3.1.0",
         "behat/web-api-extension": "~1.0@dev",
         "behat/symfony2-extension": "~2.0@dev",
         "phpbench/phpbench": "~0.10.0",
--- a/Resources/docs/mapping.rst
+++ b/Resources/docs/mapping.rst
@@ -44,7 +44,7 @@ literal value passed
 
 PropertyAccess allows you to access properties of an object by path, e.g.
 ``title``, or ``parent.title``. The expression allows you to build expressions
-which can be evaluated, e.g. ``'/this/is/' ~ object.id ~ '/a/path'``.
+which can be evaluated, e.g. ``'/this/is/' ~ object.getId() ~ '/a/path'``.
 
 Fields
 ------
@@ -88,7 +88,7 @@ language.
     <massive-search-mapping xmlns="http://massiveart.com/schema/dic/massive-search-mapping">
         <mapping class="Massive\Bundle\SearchBundle\Tests\Resources\TestBundle\Entity\Product">
             <!-- ... -->
-            <url expr="'/path/to/' ~ article.title'" />
+            <url expr="'/path/to/' ~ article.getTitle()'" />
             <!-- ... -->
         </mapping>
     </massive-search-mapping>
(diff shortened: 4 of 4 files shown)


wachterjohannes:
@benbender could you try to include `dev-feature/symfony3` in your project? the tests are running after updating the dependencies. also locally for me it seems to work on the first look.

benbender:
I tried the branch and have to report, that it fails the same way as before. It tries to access a property on my entity named after the index-name - and fails.

F.e. on a entity named event with an index named "eventfoo" i got the following exception:
````
  [Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException]
  Neither the property "eventfoo" nor one of the methods "getEventfoo()", "eventfoo()", "isEventfoo()", "hasEventfoo()", "__get()" exist and have public ac
  cess in class "AppBundle\Entity\Event".
````

Any idea on this?

**Comment by wachterjohannes** [comment]
could you post your xml and class file?

**Comment by wachterjohannes** [comment]
and also a stacktrace?

**Comment by wachterjohannes** [comment]
could it be that you define your index with `<index property="contact" />` instead `<index value="contact" />`

**Comment by benbender** [comment]
Ahh, you found the right path - it's in fact a documentation issue. I followed the documentation here:
http://massivesearchbundle.readthedocs.org/en/latest/introduction.html

And used "<index NAME="post" />"

But there seem to be another glitch. With <id property="id" /> and id defined as private property and a getter, i get the following exception. Defining $id as public works and fails only for the id:

> Fatal error:  Cannot access private property AppBundle\Entity\Post::$id in /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Component/ExpressionLanguage/Node/GetAttrNode.php on line 78
> PHP Stack trace:
> PHP   1. {main}() /Users/bb/Desktop/c4a/bin/console:0
> PHP   2. Symfony\Component\Console\Application->run() /Users/bb/Desktop/c4a/bin/console:29
> PHP   3. Symfony\Bundle\FrameworkBundle\Console\Application->doRun() /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:117
> PHP   4. Symfony\Component\Console\Application->doRun() /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:80
> PHP   5. Symfony\Component\Console\Application->doRunCommand() /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:186
> PHP   6. Symfony\Component\Console\Command\Command->run() /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:807
> PHP   7. Massive\Bundle\SearchBundle\Command\ReindexCommand->execute() /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Component/Console/Command/Command.php:256
> PHP   8. Massive\Bundle\SearchBundle\Command\ReindexCommand->reindexClass() /Users/bb/Desktop/c4a/vendor/massive/search-bundle/Command/ReindexCommand.php:149
> PHP   9. Massive\Bundle\SearchBundle\Search\SearchManager->index() /Users/bb/Desktop/c4a/vendor/massive/search-bundle/Command/ReindexCommand.php:219
> PHP  10. Massive\Bundle\SearchBundle\Search\ObjectToDocumentConverter->objectToDocument() /Users/bb/Desktop/c4a/vendor/massive/search-bundle/Search/SearchManager.php:136
> PHP  11. Massive\Bundle\SearchBundle\Search\Metadata\FieldEvaluator->getValue() /Users/bb/Desktop/c4a/vendor/massive/search-bundle/Search/ObjectToDocumentConverter.php:83
> PHP  12. Massive\Bundle\SearchBundle\Search\Metadata\FieldEvaluator->getExpressionValue() /Users/bb/Desktop/c4a/vendor/massive/search-bundle/Search/Metadata/FieldEvaluator.php:58
> PHP  13. Symfony\Component\ExpressionLanguage\ExpressionLanguage->evaluate() /Users/bb/Desktop/c4a/vendor/massive/search-bundle/Search/Metadata/FieldEvaluator.php:120
> PHP  14. Symfony\Component\ExpressionLanguage\Node\BinaryNode->evaluate() /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Component/ExpressionLanguage/ExpressionLanguage.php:70
> PHP  15. Symfony\Component\ExpressionLanguage\Node\GetAttrNode->evaluate() /Users/bb/Desktop/c4a/vendor/symfony/symfony/src/Symfony/Component/ExpressionLanguage/Node/BinaryNode.php:112
>

**Comment by wachterjohannes** [comment]
could you please post your xml i think you use `expr="object.id` to get the id instead of `property="id"`-

**Comment by benbender** [comment]
Here it is:

````<massive-search-mapping xmlns="http://massive.io/schema/dic/massive-search-mapping">

    <mapping class="AppBundle\Entity\Post">
        <index value="post" />
        <id property="id" />
        <title property="title" />
        <url expr="'/post/' ~ object.id" />
        <description property="teaser" />
        <image expr="'/assets/images/' ~ object.id" />
        <fields>
            <field name="title" type="string" />
            <field name="body" type="string" />
        </fields>
    </mapping>

</massive-search-mapping>````

But you are on the right track again. If I remove the mappings for url and image, it works as "intended". So usage of the expression language "fails" atm for sf3.

**Comment by wachterjohannes** [comment]
in a expression you have to use the function directly in your case `object.getId()`

**Comment by wachterjohannes** [comment]
there is already an issue in symfony which discuss this fact https://github.com/symfony/symfony/issues/18520

**Comment by wachterjohannes** [comment]
the rest should be valid in my oponion. so if this works for you we can merge that. /cc @danrot

**Comment by benbender** [comment]
Ahh, okay. Thanks for the hint. Changing it to the name of the method works as intended.

I would propose that you also make it clear in your docs. It isn't obvious at the moment, because f.e. ```<url expr="'/path/to/' ~ object.id" />``` is used everywhere.

Thanks again for your effort!

File: .travis.yml [hunk 1]
```diff
@@ -7,17 +7,22 @@ sudo: false
 
 php:
   - 5.5
-  - 5.6
   - 7.0
 
+env:
+  matrix:
+    - PREFER="--prefer-lowest"
+    - PREFER=""
```
dantleech: I would only do this with one version of PHP (5.5)

.travis.yml
```diff
@@ -7,17 +7,22 @@ sudo: false
...
-    - composer install
+    - composer update $PREFER
 
 script: 
     - vendor/bin/phpunit --coverage-text
     - vendor/bin/behat --suite=zend_lucene
     - vendor/bin/behat --suite=elastic
 
-env: 
-  - SYMFONY_DEPRECATIONS_HELPER=weak
```
dantleech: We still need this ?

wachterjohannes: no because it should fail if we use something deprecated of symfony 3.

.travis.yml
wachterjohannes: how can i choose in which version the matrix exists?

wachterjohannes:
@benbender i have already updated the docs in this PR. thanks for you help!

.travis.yml
wachterjohannes: i have found it (:

wachterjohannes:
@dantleech if the tests are green we can merge it. /cc @chirimoya should we create a new version?

.travis.yml
dantleech: I think something like:

```
matrix:
    include:
        - php: 5.6
          env: COMPOSER_FLAGS="--prefer-lowest"
        - php: 7
```

I think you may need to remove the `php` section in order to have 2
builds.

https://github.com/phpbench/phpbench/blob/master/.travis.yml#L28
https://github.com/symfony-cmf/routing-auto/blob/master/.travis.yml#L17

behat.yml
```diff
@@ -7,23 +7,23 @@ default:
                     adapterId: massive_search.adapter.zend_lucene
             bundle: MassiveSearchBundle
             filters:
-                tags: @zend_lucene
+                tags: '@zend_lucene'
```
danrot: I think we are always using `"` in yml files.

composer.json
```diff
@@ -10,20 +10,22 @@
     ],
     "require": {
         "jms/metadata": "~1.5",
-        "symfony/property-access": "~2.4",
+        "symfony/property-access": "~2.6||~3.0",
```
danrot: We are not supporting 2.4 anymore?

wachterjohannes: symfony 2.4 is death (: i think it would be no problem should i revert this to 2.4?

behat.yml
wachterjohannes: i have read yesterday that if your are using `'` you dont have to mask everything (except `'`)

composer.json
dantleech: s/death/dead :) We are < 1.0 so technically we *could*, but it would break peoples installations, f.e. https://packagist.org/packages/esperanto/esperanto-cms-project is tagged to `0.2` but there may be other people not using packagist that are using `~0.2` or similar.

wachterjohannes: the question here is did we want to maintain 2.6 after releasing 1.0 ? if not we should change it now.

danrot: No, you've got a good point here. Other people have to assume that stuff like this changes... but apart from that 2.6 is also dead since 01/16 😉 So I would even go to `~2.7`.

wachterjohannes:
@danrot changed to 2.7 how will we release this? 0.15 or 0.14.1?

danrot:
This change requires a `0.15.0` I'd say.

wachterjohannes:
i also think so. then we have to aware of changes until the next release.

wachterjohannes:
@danrot @dantleech ping

patkar:
backref: sulu/sulu#2106

PR #86, opened 2016-04-18
Adds mapping for the image field

### LINKED ISSUE #84: Mapping for the image-field missing in XmlDriver
I'm playing around with this nice bundle and noticed that the image-fields of my search-index were empty even if I defined them. After digging through the code, it seems to me, that the mapping is missing. 

With the following additions it works as expected:
```
diff --git a/Search/Metadata/Driver/XmlDriver.php b/Search/Metadata/Driver/XmlDriver.php
index bfc3f76..0e7f044 100644
--- a/Search/Metadata/Driver/XmlDriver.php
+++ b/Search/Metadata/Driver/XmlDriver.php
@@ -118,6 +118,7 @@ class XmlDriver extends AbstractFileDriver implements DriverInterface
             $indexMetadata->setTitleField($mapping['title']);
             $indexMetadata->setUrlField($mapping['url']);
             $indexMetadata->setDescriptionField($mapping['description']);
+            $indexMetadata->setImageUrlField($mapping['image']);

             foreach ($mapping['fields'] as $fieldName => $fieldData) {
                 $indexMetadata->addFieldMapping($fieldName, $fieldData);
@@ -148,6 +149,9 @@ class XmlDriver extends AbstractFileDriver implements DriverInterface
         $urlField = $this->getMapping($mapping, 'url');
         $indexMapping['url'] = $urlField;

+        $imageField = $this->getMapping($mapping, 'image');
+        $indexMapping['image'] = $imageField;
+
         $descriptionField = $this->getMapping($mapping, 'description');
         $indexMapping['description'] = $descriptionField;
```

Am I missing something or is this feature really missing in current master?

benbender:
Here comes the pull request, which adds the mapping for the image-field missing in the XmlDriver. 

As requested, it fixes #84.

### DIFF
--- a/Search/Metadata/Driver/XmlDriver.php
+++ b/Search/Metadata/Driver/XmlDriver.php
@@ -118,6 +118,7 @@ class XmlDriver extends AbstractFileDriver implements DriverInterface
             $indexMetadata->setTitleField($mapping['title']);
             $indexMetadata->setUrlField($mapping['url']);
             $indexMetadata->setDescriptionField($mapping['description']);
+            $indexMetadata->setImageUrlField($mapping['image']);
 
             foreach ($mapping['fields'] as $fieldName => $fieldData) {
                 $indexMetadata->addFieldMapping($fieldName, $fieldData);
@@ -150,6 +151,9 @@ class XmlDriver extends AbstractFileDriver implements DriverInterface
 
         $descriptionField = $this->getMapping($mapping, 'description');
         $indexMapping['description'] = $descriptionField;
+
+        $imageField = $this->getMapping($mapping, 'image');
+        $indexMapping['image'] = $imageField;
 
         return $indexMapping;
     }
--- a/Tests/Features/general.feature
+++ b/Tests/Features/general.feature
@@ -114,7 +114,7 @@ Feature: Search Manager
                 "description": "Hello",
                 "class": "Massive\\Bundle\\SearchBundle\\Tests\\Resources\\TestBundle\\Entity\\Car",
                 "url": "foobar",
-                "image_url": "",
+                "image_url": "foo.jpg",
                 "locale": "fr"
             }
         ]


wachterjohannes:
thanks @benbender could you please take a look into the test results. i think it should be easy to fix (: additionally please add a line for you fix into the CHANGELOG.md file in the root directory.

dantleech:
@benbender are you able to add a test, I would think just adding an image mapping [here](https://github.com/massiveart/MassiveSearchBundle/blob/develop/Tests/Features/index.feature#L43) and asserting the result would suffice.

dantleech:
@benbender (ignore last comment). Infact there already is a test for this, but it asserts that the image is empty. See [here](https://travis-ci.org/massiveart/MassiveSearchBundle/jobs/124000103#L1379).

You need only update the test to assert that image is equal to `foo.jpg` [here](https://github.com/massiveart/MassiveSearchBundle/blob/develop/Tests/Features/general.feature#L117).

Run behat as follows:

```bash
$ composer install
$ vendor/bin/behat --suite=zend_lucene
```

benbender:
Here we are. Maybe you could explain to me why this was missing? It is a feature which exists in the docs and in the tests. I thought that there was simply a faulty refactoring which forgot the field and thats why it was missing. But now it seems to me, that that feature wasn't stable anyways.

To recap: from your point of view, is it a completely new feature or a simple bugfix?

dantleech:
@benbender an oversight. We only use the image feature with the Sulu Content driver (not included here), so we never used it with the Xml Driver.

dantleech:
Squashed the commit, added the CHANGELOG and merged in ba51ba1bca4ff057c39bc552216ac36cfa292fd7.

Thanks @benbender !

benbender:
Okay, got it. Thanks for the explanation!

As a sidenote and feedback: again it's a documentation issue. The feature is documented in the official docs and it was not intended to be supported in this driver. Maybe there would be a coordinated effort needed to check the documentation against the acutal and intended featureset because in fact all of my problems were actually documentation issues :)

dantleech:
>> and it was not intended to be supported in this driver.

It was intended, but it was overlooked. Our bad.

Created following issue to review documentation:

    https://github.com/massiveart/MassiveSearchBundle/issues/87

benbender:
@dantleech Thank you!

PR #90, opened 2016-05-09
Fixed elastic search params array

wachterjohannes:
This PR fixes the config for elastisearch client constructor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG
 =========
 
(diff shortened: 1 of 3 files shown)


danrot:
@wachterjohannes Is it possible to add a test in some way? Otherwise we won't detect it another time if something is going wrong.

wachterjohannes:
@danrot how would you do that?

danrot:
I don't know, that's why I am asking 😉

PR #92, opened 2016-05-10
Fixed memory usage for reindex

wachterjohannes:
This PR introduces the cleanup after each "batch" which was reindexed. This reduces the memory usage a lot.


Command/ReindexCommand.php
```diff
@@ -219,7 +221,6 @@ private function reindexClass(
                         $this->searchManager->index($object, $locale);
                     }
                     $progress->advance();
-                    $output->write(' Mem: ' . number_format(memory_get_usage()) . 'b');
```
dantleech: Is that not useful?

wachterjohannes: i have added this to the progress-bar format which will be printed by symfony 
image

dantleech: Nice!

PR #96, opened 2016-06-16
Updated elasticsearch dependency

wachterjohannes:
This PR upgrades the elasticsearch library.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,11 +14,11 @@
         "php": ">=5.4.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.5",
+        "phpunit/phpunit": ">=4.8",
 
         "zendframework/zend-stdlib": "2.3.1 as 2.0.0rc5",
         "zendframework/zendsearch": "2.*@dev",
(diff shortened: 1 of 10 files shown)


composer.json
```diff
@@ -29,7 +29,9 @@
         "behat/web-api-extension": "~1.0@dev",
         "behat/symfony2-extension": "~2.0@dev",
         "phpbench/phpbench": "~0.10.0",
-        "phpbench/tabular": "@dev"
+        "phpbench/tabular": "@dev",
+
+        "guzzlehttp/streams": "^3.0.0"
```
danrot: Why is that needed? The changes in here don't seem to require guzzle.

wachterjohannes: if i remove this line the `--prefer-lowest` test doesnt work anymore (because of composer) https://travis-ci.org/massiveart/MassiveSearchBundle/jobs/140464875#L182

wachterjohannes: i havent found any other solution.

danrot: Hm, looks like that's a bug in the elasticsearch library? But well, looks like we don't have any other choice at the moment.

PR #97, opened 2016-06-29
Implemented sorting for zend-lucene-adapter

wachterjohannes:
This feature introduces simple sorting for zend-lucene adapter.

Additionally this PR enables empty searches for elastic-search and zend-lucene.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ CHANGELOG
 =========
 
 * dev-develop
+    * FEATURE     #97 Implemented sorting for zend-lucene-adapter
+    * FEATURE     #79 Consider limit + offset for searching
     * ENHANCEMENT #96 Updated elasticsearch dependency
     * BUGFIX      #93 Fixed composer-events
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@ CHANGELOG
 =========
 
 * dev-develop
+    * FEATURE     #97 Implemented sorting for zend-lucene-adapter
+    * FEATURE     #79 Consider limit + offset for searching
```
wachterjohannes: this entry was forgotten.

danrot:
Is that still needed? Or have you decided to go another way? /cc @chirimoya

danrot:
Based on the discussion we once had with @chirimoya I would still say that sorting is not part of the scope of this bundle. Therefore I would even go the other way and remove the sort functionality completely (Although it is a BC break, but we haven't released 1.0 anyway, so now is the chance to get rid of this)

wachterjohannes:
lets discuss this tomorrow with @chirimoya

PR #100, opened 2016-08-15
minor documentation fixes

martinsik:
I fixed a few typos and a mistakes in the documentation. However, there're still some areas that need more attention such as paginating results or writing custom convertors.

### DIFF
--- a/Resources/docs/extending.rst
+++ b/Resources/docs/extending.rst
@@ -70,7 +70,7 @@ definition.
 
     <service id="massive_search.metadata.driver.xml" class="%massive_search.metadata.driver.xml.class%">
         <argument type="service" id="massive_search.metadata.file_locator" />
-        <tag type="massive_search.metadata.driver" />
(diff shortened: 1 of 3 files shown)


wachterjohannes:
@martinsik thanks for fixing this documentation bugs. may i ask you where you use this project?

martinsik:
Hey, I'm implementing it into our proprietary CMS in the company I work right now. Since I'm the only one working on it, this bundle saved me a lot of time and it does almost all I need. I noticed this bundle is also used in SULU CMS and now I see it's made by the same company :).

Btw, I'd like to hear your reasoning in MASSIVE ART that you choose to made the SULU CMS open source because that's basically the same situation we're dealing with now in PRIA (www.pria.cz). We're creating mostly typical "digital agency" stuff so smaller projects but highly customizable, that's why we choose to write our of CMS based on Symfony and not use any already existing one.

PR #102, opened 2016-08-28
Added .gitkeep

### LINKED ISSUE #101: Test suites are dependent on the order they're executed
When you clone the repo and try to run only the Zend Lucene tests (`vendor/bin/behat --suite=zend_lucene`) you end up with many same errors:

    There is no search mapping for object with class "Massive\Bundle\SearchBundle\Tests\Resources\TestBundle\Entity\Product" (Massive\Bundle\SearchBundle\Search\Exception\MetadataNotFoundException)

This happens because [MassiveSearchExtension on line 123 ](https://github.com/massiveart/MassiveSearchBundle/blob/develop/DependencyInjection/MassiveSearchExtension.php#L123) checks whether the `massive-search` exists, which doesn't. Also, directory `Entity` doesn't exist.

The traviscl tests run fine because they're executed in order where `vendor/bin/phpunit --coverage-text` is run first. As a side effect it creates also `massive-search` and `Entity` directories and doesn't remove them when it finishes (which it probably should).

To see how it fails run:

```
$ git clone https://github.com/massiveart/MassiveSearchBundle.git
$ cd MassiveSearchBundle
$ composer install
$ vendor/bin/behat --suite=zend_lucene
```

To execute the tests properly run:

```
$ git clone https://github.com/massiveart/MassiveSearchBundle.git
$ cd MassiveSearchBundle
$ composer install
$ mkdir Tests/Resources/TestBundle/Resources/config/massive-search
$ mkdir Tests/Resources/TestBundle/Entity
$ vendor/bin/behat --suite=zend_lucene
```

martinsik:


wachterjohannes:
@martinsik is this the fix for https://github.com/massiveart/MassiveSearchBundle/issues/101 ?

martinsik:
@wachterjohannes yep

wachterjohannes:
@martinsik thanks for contribution.

PR #107, opened 2017-05-09
Added encode and decode field-name to elasticsearch adapter

wachterjohannes:
Fixes https://github.com/sulu/sulu/issues/3274

The dot is a reserved character in elasticsearch (besides `:`, `+` and `-`) - so we have decided to replace it with something else (on index and search time).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG
 =========
 
+* dev-master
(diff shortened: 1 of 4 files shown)


wachterjohannes:
@danrot now the tests running.

danrot:
Code LGTM, @dominikmatt can you test it?

PR #108, opened 2017-06-14
No size when limit empty

Bacis:
Once the limit is not set do not use size paramater in elasticsearch query.

https://github.com/sulu/sulu/issues/3406

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 CHANGELOG
 =========
 
+* 0.16.2 (2017-06-14)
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@Bacis thanks for the bugfix :) could you please add also the CHANGELOG.md line?

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 CHANGELOG
 =========
 
+* 0.16.2 (2017-06-14)
```
wachterjohannes: can you remove the date (: i am nor sure if this will be relased today. we will update this when doing! use `dev-master` as headline

PR #110, opened 2017-10-13
Highlighting of search hits in searched fields

adsc-cloudtec:
This adds the capability to specify an array of fields in the SearchQuery, which will be used to find context around search result "hits" (which seem to be called "highlight" in the searching world). The ElasticSearchAdapter then modifies the query to include configuration to return those highlights, and puts the returned highlight data from Elasticsearch into the SearchDocument.

Example usage looks like this:
```
$searchManager = $this->get('massive_search.search_manager');
$hits = $searchManager->createSearch($query)->index($index)->highlight(['title', 'summary', 'content'])->execute();
foreach ($hits as $hit) {
   $document = $hit->getDocument();
   $highlight = $document->getHighlight();
   // Do something with the highlight data
}
```

The highlight property on the document is an array and looks something like this:
```
[
   "fieldname1" => [
      "context around hit1 in fieldname1's content",
      "context around hit2 in fieldname1's content",
      ...
   ],
   "fieldname2" => [
      "context around hit1 in fieldname2's content",
      "context around hit2 in fieldname2's content",
      ...
   ],
   ...
]
```
If no hit happened in a field that was specified for highlighting, it won't be present in the result document's highlight array. If no hit happened in any of the fields, the highlight array will be empty.

Zend Lucene adapter also has capability for highlights, but this is not implemented in this PR.

### DIFF
--- a/Search/SearchQueryBuilder.php
+++ b/Search/SearchQueryBuilder.php
@@ -118,7 +118,8 @@ class SearchQueryBuilder
     }
 
     /**
-     * Set which fields should be highlighted with search hits
+     * Set which fields should be highlighted with search hits.
+     *
      * @param array $highlight
      *
      * @return SearchQueryBuilder
--- a/Search/SearchQueryBuilder.php
+++ b/Search/SearchQueryBuilder.php
@@ -118,6 +118,19 @@ class SearchQueryBuilder
     }
 
     /**
+     * Set which fields should be highlighted with search hits
+     * @param array $highlight
+     *
+     * @return SearchQueryBuilder
+     */
+    public function highlight($highlight)
+    {
+        $this->searchQuery->setHighlightFields($highlight);
+
+        return $this;
+    }
+
+    /**
      * Execute the search.
      *
      * @return SearchResult
--- a/Search/Adapter/ElasticSearchAdapter.php
+++ b/Search/Adapter/ElasticSearchAdapter.php
@@ -165,7 +165,16 @@ class ElasticSearchAdapter implements AdapterInterface
         if (!empty($searchQuery->getLimit())) {
             $params['body']['size'] = $searchQuery->getLimit();
         }
+        if (!empty($searchQuery->getHighlightFields())) {
+            $params['body']['highlight'] = [
+                "require_field_match" => false,
+                "fields" => []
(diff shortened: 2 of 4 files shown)


Search/SearchQueryBuilder.php
```diff
@@ -118,6 +118,19 @@ public function setOffset($offset)
     }
 
     /**
+     * Set which fields should be highlighted with search hits
```
wachterjohannes: put a point at the end of the sentence and a line-break at the end of the line.

wachterjohannes:
i will take a look at the tests today https://travis-ci.org/massiveart/MassiveSearchBundle/builds/243544556

adsc-cloudtec:
I hope I can also look at the zend-search part, but at the moment there are more pressing matters I must attend to.

almare:
I stumbled while I searching for the highlight feature about this topic. Why is this topic closed? I ask my self how other solve their website search without this es feature? Is it possible to get this changes merged ?

danrot:
@almare The PR was closed because there was no progress for 2 years, and the code was only working for elastic search, not for the zend lucene adapter. If you find a generic solution that works for both, we might consider merging it.

And is my assumption right, that this should make sure that some part of a shortened text contains the higlighted word?

almare:
@danrot Your assumption is absolutely right. It's more like a google search instead of a plain text dump which has nothing to do with the search. I can understand your argument to look for a generic search, but isn't it better to have one working solution instead of nothing? 

ES is a hard requirement for the ArticleBundle, so users which use this bundle will directly get improved search results, others are free to add the missing LuceneAdapter feature. Correct me if I'm wrong.

danrot:
At the very least I would like to know if this would also be possible for Zend Lucene, and if it is easy to implement I would argue we should do that immediately, because otherwise it will never be done...

almare:
This sound great :-) Honestly I must say I have never worked with Zend Lucene and I am wondering I don't find any current documentation on google. 
I am also not sure if the behaviour is the same. In ES I just receive the highlighted result snippets or better excerpts. I don't know if Lucene just add a highlighting tag to the search term in the search result or if a an excerpt will created with the highlighted term like ES did.

PR #112, opened 2017-10-16
Added a section about paging the search results

adsc-cloudtec:
This small PR adds a short section in the documentation about how to set up the SearchQuery to control paging of results.


trickreich:
@wachterjohannes Is ok for me

PR #115, opened 2018-03-16
Updated Elastic Search dev-Dependency

### LINKED ISSUE #114: Elastic Search Dependency is EOL
The latest Elastic Search release supported by elastic/elasticsearch-php@2.1 [1] is Elastic Search 2.4.x to which all support ended on 28.02.2018 [2].

It looks like the elastic feature set used by MassiveSearchBundle is fully supported by Elastic Search 6 and therefore elastic/elasticsearch-php should be upgraded to the 6.0 branch.

Links:
------

[1] https://github.com/elastic/elasticsearch-php/tree/master#version-matrix
[2] https://www.elastic.co/support/eol

pascalvb:
Fixes #114

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
 
         "zendframework/zend-stdlib": "2.3.1 as 2.0.0rc5",
         "zendframework/zendsearch": "2.*@dev",
-        "elasticsearch/elasticsearch": "^2.1",
+        "elasticsearch/elasticsearch": "^2.1||^6.0",
         "symfony-cmf/testing": "^1.3",
         "symfony/symfony": "~2.7||~3.0",
         "matthiasnoback/symfony-dependency-injection-test": "0.7.*",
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@pascalvb thanks for this thought and PR but I think we should to support elasticsearch `^2.0` and elasticsearch `^6.0` do you think its possible?

pascalvb:
hi @wachterjohannes, sorry for the delay. I changed the dependency to support both ```^2.0``` and ```^6.0```, this shouldn't be an issue that way around.

composer.json
```diff
@@ -18,7 +18,7 @@
 
         "zendframework/zend-stdlib": "2.3.1 as 2.0.0rc5",
         "zendframework/zendsearch": "2.*@dev",
-        "elasticsearch/elasticsearch": "^2.1",
+        "elasticsearch/elasticsearch": "^2.0|^6.0",
```
wachterjohannes: go with `^2.1|^6.0`

wachterjohannes:
@pascalvb could you update this branch with current develop?

PR #118, opened 2018-05-02
Minor fix : elasticsearch _type

wimvds:
When looking at the ElasticSearch indexes created with Sulu, I noticed that the _type field looked like this :
```      
{
        "_index" : "massive_page-fr-i18n",
        "_type" : "ulu_Bundle_ContentBundle_Document_HomeDocument",
        "_id" : "cc6738ef-7fe2-4707-ba3f-0e2dac854700",
        "_score" : 1.0,
        "_source" : {
          "excerptTitle" : "",
          ...
        }
} 
```

As you can see, the _type is missing the first letter. This PR fixes this (by stripping any preceding _ after converting the \ in the class names to _)...

### DIFF
--- a/Search/Adapter/ElasticSearchAdapter.php
+++ b/Search/Adapter/ElasticSearchAdapter.php
@@ -306,7 +306,7 @@ class ElasticSearchAdapter implements AdapterInterface
             return 'massive_undefined';
         }
 
-        return substr(str_replace('\\', '_', $class), 1);
+        return ltrim(str_replace('\\', '_', $class), '_');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


wachterjohannes:
@wimvds thanks for the contribution - but it seems to be a BC break isnt it? we should at least write it into the UPGRADE.md file

wimvds:
Ow, yeah. It does require one to recreate the index (though I suspect the old entries will just be ignored). What version should I add in the UPGRADE.md (or will you add the version later on)?

wachterjohannes:
@wimvds you can simply write `dev-develop` as the version

PR #119, opened 2018-05-16
Change default directory to var folder

alexander-schranz:


wachterjohannes:
@alexander-schranz i would say go for it :) but the tests are failing

alexander-schranz:
to avoid bc break in this repository I will prepend configuration for sulu there

PR #120, opened 2018-05-28
Update queryBuilder with result of custom repository method

pierredup:
When using a custom repository method, the result of the method is not currently used when returning a new QueryBuilder instance, which causes the custom query to be ignored. This patch updated the method to mimic what is done in the `getCount`


pierredup:
Actually, going through the code again, I see it's more of a documentation issue. The docs are not clear that a `QueryBuilder` instance is passed to the method that should be used to construct the query.
I've opened #121 to rather address the issue in the docs.

PR #121, opened 2018-05-29
Add a note about using the QueryBuilder when using a custom repository method

pierredup:
The documentation is not clear about how to use a custom repository method, so I would like to add a note to indicate that the passed `QueryBuilder` instance should be used when constructing a query


wachterjohannes:
:+1: thanks for this!

PR #127, opened 2019-02-28
Custom repository method bugfix

thorne51:
When specifying a custom repository method for reindex which does not return anything as per the documentation, the `DoctrineOrmProvider::getCount()` method rebuilds the query builder using the entity manager. This patch merely mitigates the behaviour so that custom repository methods that adhere to the documentation specification does not produce unexpected results.


wachterjohannes:
@thorne51 thanks for the contribution - your totally right this was an issue in the code-base.

PR #132, opened 2019-09-19
Update jms/metadata dependency

alexander-schranz:


Tests/Unit/Search/Metadata/Driver/XmlDriverTest.php
```diff
@@ -49,11 +49,12 @@ public function setUp()
 
         $this->reflectionClass = $this->prophesize(\ReflectionClass::class);
         $this->reflectionClass->getName()->willReturn('Sulu\Bundle\ExampleBundle\Entity\Example');
+        $this->reflectionClass->__toString()->willReturn('what the hell');
```
danrot: Is that on purpose? Looks like something you just put there temporarily.

UPGRADE.md
```diff
@@ -1,5 +1,16 @@
...
+## dev-develop
+
+## JMS/Metadata package update to 2.0
+
+To update JMS/Metadata to 2.0 the following public functions were removed:
+
+ - Massive\Bundle\SearchBundle\Search\Metadata\Driver::loadMetadataFromFile
+ - Massive\Bundle\SearchBundle\Search\Metadata\Driver::getExtension
+
+In this case we alse needed to increase the required `php` version to `7.2`.
```
danrot: `In this case` does not feel right here... I would write something like "Because of the update we had to increase the required php version to 7.2"

composer.json
```diff
@@ -9,18 +9,30 @@
...
         "symfony-cmf/testing": "^1.3 || ^2.1",
-        "symfony/symfony": "~2.7 || ~3.0 || ~4.0",
+        "symfony/filesystem": "~3.4 || ~4.0",
+        "symfony/finder": "~3.4 || ~4.0",
+        "symfony/http-kernel": "~3.4 || ~4.0",
+        "symfony/framework-bundle": "~3.4 || ~4.0",
+        "symfony/form": "~3.4 || ~4.0",
+        "symfony/monolog-bridge": "~3.4 || ~4.0",
+        "symfony/validator": "~3.4 || ~4.0",
+        "symfony/templating": "~3.4 || ~4.0",
```
danrot: Since we tag that as 2.0, shouldn't we also support only the latest symfony version here?

Tests/Unit/Search/Metadata/Driver/XmlDriverTest.php
alexander-schranz: ups will change the text but something in prophecy itself is do a __toString now.

composer.json
alexander-schranz: Will upgrade the dependencies.

alexander-schranz:
@danrot fixed

PR #136, opened 2020-01-30
Add support for elasticsearch 6 and 7

alexander-schranz:
Add support for elasticsearch 6.

Currently it will fail because elasticsearch 6 does not longer support multiple types per index. So if you now use elasticsearch 6 we will add a custom document type field and set the type to the index_name.

### DIFF
--- a/Search/Adapter/ElasticSearchAdapter.php
+++ b/Search/Adapter/ElasticSearchAdapter.php
@@ -43,6 +43,8 @@ class ElasticSearchAdapter implements AdapterInterface
 
     const IMAGE_URL = '__image_url';
 
+    const DOCUMENT_TYPE = '__document_type';
+
     /**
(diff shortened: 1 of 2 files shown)


Search/Adapter/ElasticSearchAdapter.php
```diff
@@ -117,17 +121,22 @@ public function index(Document $document, $indexName)
...
 
         $params = [
             'id' => $document->getId(),
+            'type' => $indexName,
             'index' => $indexName,
-            'type' => $this->documentToType($document),
             'body' => $fields,
         ];
 
+        if (version_compare($this->version, '6.0', '<')) {
```
wachterjohannes: add comment

alexander-schranz: added

PR #139, opened 2020-02-23
Remove the behat/symfony2-extension and behat/web-api-extension

alexander-schranz:


danrot:
@alexander-schranz It would be nice if you would add why this has been removed in the description for future reference.

PR #140, opened 2020-04-16
Fix the data presented in the exception while validating the query

gisostallenberg:
Before the exception reports all used indexes not found, even if they exist.


alexander-schranz:
And could you rebase it to `release/1.x`?

gisostallenberg:
@alexander-schranz changed the base and fixed the tests

alexander-schranz:
@gisostallenberg Thank you for your contribution 👍

PR #144, opened 2020-04-23
Add namespaces for metadata related to the kernel namespace

gisostallenberg:


DependencyInjection/Compiler/MetadataFileLocatorPass.php
```diff
@@ -0,0 +1,49 @@
...
+        if ('App\\Kernel' === $kernelClass) {
+            return; // App paths already set in MassiveSearchExtension::loadMetadata
+        }
+
+        $projectNamespace = substr($kernelClass, 0, strrpos($kernelClass, '\\'));
+        $kernelProjectDir = $container->getParameter('kernel.project_dir');
+        $fileLocator = $container->getDefinition('massive_search.metadata.file_locator');
+        $metadataPaths = $fileLocator->getArgument(0);
+        foreach (['Entity', 'Document', 'Model'] as $entityNamespace) {
+            if (!file_exists($kernelProjectDir . '/src/' . $entityNamespace)) {
```
alexander-schranz, line 39: If we doing this I would not build ontop not using project_dir to find the documents as we don't know if its in the `src` or another folder like `src/Sulu` or something, because I thought thats currently the issue you have, so I would use here reflection to get to the KernelDirectory:

```
$kernelClass = $container->getDefinition('kernel')->getClass();
$reflectionClass = new \ReflectionClass($kernelClass);
$kernelProjectDir = $container->getParameter('kernel.project_dir');
$kernelDirectory = dirname($reflectionClass->getFileName());

        foreach (['Entity', 'Document', 'Model'] as $entityNamespace) {
            if (!file_exists($kernelDirectory . $entityNamespace)) {
                continue;
            }

            $namespace = 'App\\' . $entityNamespace;
            $metadataPaths[$namespace] = $kernelProjectDir . '/config/massive-search';
        }
```

Also would move the logic to the Extension there you should also have access to all data which is needed. And would just do an `array_unique` there at the end so it doesn't depend of `App\Kernel`.

gisostallenberg: I do not know how to get the kernel in the Extension.

gisostallenberg: Updated so it uses the location of the kernel gotten from reflection

gisostallenberg: In the bundle (and tests) there is no kernel class...

alexander-schranz: Ok was sure the kernel was also available in the extension but you are right. 
Then we should check at the top also with `$container->hasDefinition('kernel')`.

alexander-schranz:
@gisostallenberg thank you! Is released as 2.2.2.

PR #145, opened 2020-05-25
URL and Title should be in the index

ampaze:
This makes it possible to give the title a boost via the search query. 

As discussed here https://github.com/sulu/sulu/issues/5290

### DIFF
--- a/Search/Adapter/ZendLuceneAdapter.php
+++ b/Search/Adapter/ZendLuceneAdapter.php
@@ -162,8 +162,8 @@ class ZendLuceneAdapter implements AdapterInterface
         $luceneDocument->addField(
             Lucene\Document\Field::unStored(self::AGGREGATED_INDEXED_CONTENT, implode(' ', $aggregateValues))
         );
-        $luceneDocument->addField(Lucene\Document\Field::unIndexed(self::URL_FIELDNAME, $document->getUrl()));
-        $luceneDocument->addField(Lucene\Document\Field::unIndexed(self::TITLE_FIELDNAME, $document->getTitle()));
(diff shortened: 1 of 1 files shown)


alexander-schranz:
@ampaze is this PR needed if we implement something like https://github.com/sulu/sulu/pull/5304/files to mark which fields should be indexed?

ampaze:
It would of course be better to be able to mark the fields like in https://github.com/sulu/sulu/pull/5304/files, but this requires more work.

Although I don't think there is any negative consequences in just alway indexing it.

Search/Adapter/ZendLuceneAdapter.php
```diff
@@ -162,8 +162,8 @@ public function index(Document $document, $indexName)
         $luceneDocument->addField(
             Lucene\Document\Field::unStored(self::AGGREGATED_INDEXED_CONTENT, implode(' ', $aggregateValues))
         );
-        $luceneDocument->addField(Lucene\Document\Field::unIndexed(self::URL_FIELDNAME, $document->getUrl()));
-        $luceneDocument->addField(Lucene\Document\Field::unIndexed(self::TITLE_FIELDNAME, $document->getTitle()));
+        $luceneDocument->addField(Lucene\Document\Field::text(self::URL_FIELDNAME, $document->getUrl()));
+        $luceneDocument->addField(Lucene\Document\Field::text(self::TITLE_FIELDNAME, $document->getTitle()));
```
danrot, line 166: @chirimoya I think you have the most experience with running Zend Lucene in a production environment from us. Do you think that has a big performance impact?

ampaze:
I am closing this in favor of https://github.com/sulu/sulu/pull/5304. 

Then use title instead of __title, which is a better solution anyway.

PR #148, opened 2020-06-16
Extract search in some more methods to ease extending

gisostallenberg:


alexander-schranz:
Can you do the same for the LuceneAdapter?

gisostallenberg:
> Can you do the same for the LuceneAdapter?

I could, but because of the inner workings it is not really possible to have the same method signatures.  Is that ok?

alexander-schranz:
@gisostallenberg you mean the method will  have other arguments? Yes thats okay.

gisostallenberg:
@alexander-schranz I think this ok this way.

PR #150, opened 2020-08-25
Add possibility to deindex object in specific locale

luca-rath:


luca-rath:
@alexander-schranz Until now, when calling `$searchManager->deindex($object)`, `$object` was being deindexed in all locales. I'm pretty sure that this is a bug, but can you verify that?

alexander-schranz:
@luca-rath don't think its a bug. Looks for me its currently not implemented that you can only deindex a specific locale. I think we can not change the default behaviour as if you remove currently a entity and call deindex you expect that it is removed in all locales not only the locale you provided. 

If we want to provide deindex by locale we would need to add it to the `deindex` function e.g.:

```php
public function deindex($object, $locale = null)
```

For bc we can only be added to the SearchManager not Interface.

Search/SearchManagerInterface.php
```diff
@@ -57,8 +57,9 @@ public function index($object);
      * its corresponding index.
      *
      * @param Document|object $object
+     * @param string|null $locale
      */
-    public function deindex($object);
+    public function deindex($object/*, string $locale = null*/);
```
luca-rath, line 62: @alexander-schranz is it ok like this?

alexander-schranz: 👍

Search/SearchManager.php
```diff
@@ -126,7 +126,7 @@ private function getMetadataForDocument(Document $document)
     /**
      * {@inheritdoc}
      */
-    public function deindex($object)
+    public function deindex($object, string $locale = null)
```
alexander-schranz, line 129: It looks like if somebody extend it it still errors also when its optional so we need to go this way here with this way:

```php
public function deindex($object/*, string $locale = null*/)
{
     $locale = func_num_args() >= 2 ? func_get_arg(1) : null;
```

alexander-schranz (changes requested):

Search/SearchManager.php
luca-rath: Unfortunately phpstan still fails, if you call `deindex()` with 2 parameters. Do you have an idea how to fix that?

~~~
Method
Massive\Bundle\SearchBundle\Search\SearchManagerInterface::deindex()
invoked with 2 parameters, 1 required.
~~~

alexander-schranz: Sure that error need to added to ignores. Not sure if there is any phpstan plugin or opitons to support things like `/*, string $locale = null*/`.

luca-rath: If i would call the `deindex()` method directly on the SearchManager implementation, phpstan wouldn't fail, because it recognizes the `func_num_args()` in the code. But since we're calling the method on the interface, it can't check any code for `func_num_args()` and therefore will show the error

alexander-schranz: But maybe its possible to fake it with phpdocs by adding on the class/interface `@method deIndex ...` phpdoc annotation

luca-rath: @alexander-schranz Can we merge that for now?

alexander-schranz: That the new optional parameter is correctly detected you would need to typehint in the code you use it the `SearchManagerInterface` as `SearchManager` e.g.:

```
	/**
	 * SearchManager
	 */
	private $searchManager;

	/**
	 * @param SearchManager $searchManager
	 */
    public function __construct(SearchManagerInterface $searchManager)
	{
		$this->searchManager = $searchManager;
	}

	public function someFunction(object $object): void
	{
		$this->searchManager->deindex($object, 'de');
	}
```

alexander-schranz (approved):

Search/SearchManager.php
luca-rath: yes 👍

luca-rath:
thanks 🙂

PR #151, opened 2020-09-09
Fix ReindexCommand for translated objects

luca-rath:


luca-rath:
ping @alexander-schranz

Command/ReindexCommand.php
```diff
@@ -219,11 +219,13 @@ private function reindexClass(
 
                 try {
                     foreach ($locales as $locale) {
+                        $translatedObject = $object;
+
                         if (null !== $locale) {
-                            $object = $provider->translateObject($object, $locale);
+                            $translatedObject = $provider->translateObject($object, $locale);
```
luca-rath, line 225: Before, the `$object` variable has always been overridden by this call.

alexander-schranz:
@luca-rath this should target the `1.x` branch as a bugfix.

PR #153, opened 2021-01-14
Update github actions for 2.x

luca-rath:


alexander-schranz (approved):
@luca-rath fixed the prefer lowest tests

luca-rath:
thank you! 🙂

.github/workflows/test-application.yaml
```diff
@@ -20,16 +20,19 @@ jobs:
             matrix:
```
niklasnatter, line 20: should we add a `php-cs-fixer` step somewhere?
```
            -   name: Run php-cs-fixer
                if: ${{ matrix.php-cs-fixer }}
                run: |
                    composer global require friendsofphp/php-cs-fixer --prefer-dist --no-interaction
                    GLOBAL_BIN_DIR=$(composer global config bin-dir --absolute --quiet)
                    $GLOBAL_BIN_DIR/php-cs-fixer fix --dry-run --diff
```
https://github.com/sulu/SuluArticleBundle/blob/1.2/.github/workflows/test-application.yaml#L120-L125

luca-rath: Shouldn't we then require `php-cs-fixer` as dev dependency?

niklasnatter: same question for the MassiveBuildBundle (https://github.com/massiveart/MassiveBuildBundle/pull/21) that is already merged 🙂  I think we used `PrettyCI` in these repos, right?

niklasnatter: adding it as dev dependency is okay for me too if it does not cause any problems 🙂  we did not do that in older branches of other repos because `friendsofphp/php-cs-fixer` is incompatible with Symfony 2.8

PR #154, opened 2021-01-18
Execute php-cs-fixer in test-application workflow

niklasnatter:


alexander-schranz:
Would remove `.styleci.yml` then.

alexander-schranz (approved):

PR #158, opened 2021-05-10
Fix override entities by exclude mapped superclass from indexing

alexander-schranz:


Search/Reindex/Provider/DoctrineOrmProvider.php
```diff
@@ -52,6 +52,9 @@ public function getClassFqns()
         $classFqns = [];
 
         foreach ($metadataFactory->getAllMetadata() as $classMetadata) {
+            if ($classMetadata->isMappedSuperclass) {
```
niklasnatter, line 55: maybe add a test for this? 😇

alexander-schranz: done!

PR #162, opened 2021-11-19
Fix compatibility to jms metadata 2-5-2

alexander-schranz:
The changes from [2.5.1...2.5.2 from jms metadata](https://github.com/schmittjoh/metadata/compare/2.5.1...2.5.2) did cause a segmentation fault when serializing the metadata. To avoid this we need to change how we serialize the data.

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -21,36 +21,48 @@ jobs:
             matrix:
                 include:
                     - php-version: '7.2'
+                      elasticsearch-version: '2.4.6'
+                      elasticsearch-package-version: '^2.1'
                       minimum-stability: 'stable'
                       dependency-versions: 'lowest'
                       tools: 'composer:v1'
                       php-cs-fixer: false
 
                     - php-version: '7.3'
+                      elasticsearch-version: '5.6'
+                      elasticsearch-package-version: '^5.5'
                       minimum-stability: 'stable'
                       dependency-versions: 'highest'
                       tools: 'composer:v2'
                       php-cs-fixer: false
 
                     - php-version: '7.4'
+                      elasticsearch-version: '5.6'
+                      elasticsearch-package-version: '^5.5'
                       minimum-stability: 'stable'
                       dependency-versions: 'highest'
                       tools: 'composer:v2'
                       php-cs-fixer: true
 
                     - php-version: '8.0'
+                      elasticsearch-version: '7.11'
+                      elasticsearch-package-version: '^7.11'
                       minimum-stability: 'stable'
                       dependency-versions: 'highest'
                       tools: 'composer:v2'
                       php-cs-fixer: false
 
                     - php-version: '8.1'
+                      elasticsearch-version: '7.11'
+                      elasticsearch-package-version: '^7.11'
                       minimum-stability: 'stable'
                       dependency-versions: 'highest'
                       tools: 'composer:v2'
                       php-cs-fixer: false
 
                     - php-version: '8.1'
+                      elasticsearch-version: '7.11'
+                      elasticsearch-package-version: '^7.11'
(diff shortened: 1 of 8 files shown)


goetas:
`\serialize($this->indexMetadatas)` is the problem in your current `serialize` code. 


`$this->indexMetadatas` contains a reference to the current class metadata...

see 

```php
 $indexMetadata->setName($this->name);
$indexMetadata->setClassMetadata($this);
```
since that is done in a different serialization round, the ciricular reference detection does not work.... TBH i wonder why it was working before.

you have two options imo:

1) make the serialize method as follows

```php
    public function serialize()
    {
        $data = parent::serialize();

        return \serialize([$data, $this->indexMetadatas, $this->repositoryMethod]);
    }
```
2) implement the serialization interface in `\Massive\Bundle\SearchBundle\Search\Metadata\IndexMetadata` and make sure to not serialize the class `$classMetadata` prop

alexander-schranz:
@goetas Thanks for your support and help here, really appricate that 🙏 . Will give it a try here :) and check it tomorrow. Wish you a good weekend!

Search/Metadata/ClassMetadata.php
```diff
@@ -78,18 +78,41 @@ public function getIndexMetadata($contextName)
         return $this->indexMetadatas[$contextName];
     }
 
-    public function serialize()
+    public function __serialize(): array
+    {
+        $data = [
```
alexander-schranz, line 83: calling `parent::__serialize` does end in a segmentation fault so I reimplemented the logic here.

Search/Metadata/ClassMetadata.php
```diff
@@ -78,18 +78,41 @@ public function getIndexMetadata($contextName)
...
+        $this->indexMetadatas = $indexMetadata;
 
-        return \serialize([$data, \serialize($this->indexMetadatas), $this->repositoryMethod]);
+        [
+            $this->name,
+            $this->methodMetadata,
+            $this->propertyMetadata,
+            $this->fileResources,
+            $this->createdAt,
+        ] = $data;
```
alexander-schranz, line 105: same as above calling `parent::__unserialize` didn't work like expected so we did reimplment the parent method call.

alexander-schranz (commented):
A workaround for currently is not calling `parrent::__serialize` or `parent::__unserialize` method because of a strange Segmentation Fault. Avoiding calling it seems currently fix the incompatibility.

alexander-schranz:
@goetas I copied now the implementation of the parent method directly to our methods. This seems to work now. Could not find another solution. Do you now if there are any planes that the current ClassMetadata will get additional properties?

.github/workflows/test-application.yaml
```diff
@@ -78,8 +110,12 @@ jobs:
...
-                  vendor/bin/behat --suite=zend_lucene
-                  vendor/bin/behat --suite=elastic
+            - name: Execute unit tests
+              run: vendor/bin/simple-phpunit
+
+            - name: Execute zend lucene behat tests
+              run: vendor/bin/behat --suite=zend_lucene
+
+            - name: Execute elastic behat tests
+              if: ${{ matrix.php-version < '8.0' }} # requires fixing elastic tests on 7.11 (sorting)
```
alexander-schranz, line 120: Currently could not find a way to get the Elastic 7 tests running. Sorting would require some changes. And in the CI additional tests failing where I don't get locally.

goetas:
TBH i do not understand what did you do with those `__serialize` and `__unserialize` methods. I've published https://github.com/massiveart/MassiveSearchBundle/pull/163 . it should do what had to be done

alexander-schranz:
replaced by #163

alexander-schranz:
replaced by #163 @goetas Thank you for your help with this issue!

PR #168, opened 2022-10-20
Store block values in one field

thomasduenser:
## Problem

When at least one property inside a block (or nested block) had a search tag, a new field for each of the properties was created.

With the ElasticSearch adapter, this could lead to following error:
```
Limit of total fields [1000] has been exceeded
```

## Solution

This Pull Request collects only the tagged properties inside blocks (or nested blocks) and stores these values in one field called `contentBlocks`.

## Solves following issue
https://github.com/sulu/sulu/issues/6501

## Related PR

https://github.com/sulu/sulu/pull/6868

### DIFF
--- a/Search/ObjectToDocumentConverter.php
+++ b/Search/ObjectToDocumentConverter.php
@@ -131,7 +131,7 @@ class ObjectToDocumentConverter
      * @param mixed $object
      * @param array $fieldMapping
      * @param string[] $blockValues
-     * @param string $prefix
+     * @param string $prefix Prefix the document field name (used when called recursively)
      *
      * @throws \InvalidArgumentException
      */
@@ -182,7 +182,7 @@ class ObjectToDocumentConverter
             /** @var FieldInterface $mappingField */
             $mappingField = $mapping['field'];
             $condition = method_exists($mappingField, 'getCondition') ? $mappingField->getCondition() : null;
-            $validField = $condition ? $this->fieldEvaluator->evaluateCondition($object, $condition) : false;
+            $validField = $condition ? $this->fieldEvaluator->evaluateCondition($object, $condition) : true;
 
             if (false === $validField) {
                 continue;
@@ -223,7 +223,7 @@ class ObjectToDocumentConverter
                 continue;
             }
 
-            if ('complex' !== $mapping['type'] && $value) {
+            if ('complex' !== $mapping['type']) {
                 if ($isBlockScope && $value && Field::TYPE_STRING === $type) {
                     $blockValues[$prefix . $fieldName][] = $value;
                 } elseif (!$isBlockScope) {
@@ -233,15 +233,13 @@ class ObjectToDocumentConverter
                 continue;
             }
 
-            if ($value) {
-                foreach ($value as $key => $itemValue) {
-                    $itemType = $mapping['type'];
+            foreach ($value as $key => $itemValue) {
+                $itemType = $mapping['type'];
 
-                    if ($isBlockScope && $itemValue && Field::TYPE_STRING === $itemType) {
-                        $blockValues[$prefix . $fieldName . $key][] = $itemValue;
-                    } elseif (!$isBlockScope) {
-                        $this->addDocumentField($document, $fieldName . $key, $itemValue, $mapping, $itemType);
-                    }
+                if ($isBlockScope && $itemValue && Field::TYPE_STRING === $itemType) {
+                    $blockValues[$prefix . $fieldName . $key][] = $itemValue;
+                } elseif (!$isBlockScope) {
+                    $this->addDocumentField($document, $fieldName . $key, $itemValue, $mapping, $itemType);
                 }
             }
         }
--- a/Search/ObjectToDocumentConverter.php
+++ b/Search/ObjectToDocumentConverter.php
@@ -213,7 +213,7 @@ class ObjectToDocumentConverter
             if ('complex' !== $mapping['type']) {
                 if ($isBlockScope && $value && Field::TYPE_STRING === $type) {
                     $this->blockValues[] = strip_tags($value);
-                } elseif ($value) {
+                } elseif (!$isBlockScope) {
                     $this->addDocumentField($document, $fieldName, $value, $mapping, $type);
                 }
 
--- a/Search/ObjectToDocumentConverter.php
+++ b/Search/ObjectToDocumentConverter.php
@@ -120,7 +120,7 @@ class ObjectToDocumentConverter
 
         // Adds the merged data of each content-block (even nested-blocks) to the document.
         if (0 < count($this->blockValues)) {
-            $mapping = $this->addDefaultMappingOptions();
+            $mapping = $this->addMappingOptions();
             $blockValues = implode(' ', $this->blockValues);
             $this->addDocumentField($document, 'contentBlocks', $blockValues, $mapping, Field::TYPE_STRING);
         }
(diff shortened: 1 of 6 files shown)


Search/ObjectToDocumentConverter.php
```diff
@@ -35,6 +35,11 @@ class ObjectToDocumentConverter
      */
     private $converterManager;
 
+    /**
+     * @var string[]
+     */
+    private $blockValues = [];
```
alexander-schranz: Services should be stateless.

Search/ObjectToDocumentConverter.php
```diff
@@ -226,29 +211,80 @@ private function populateDocument($document, $object, $fieldMapping, $prefix = '
...
-                        $prefix . $fieldName,
-                        $value,
-                        $type,
-                        $mapping['stored'],
-                        $mapping['indexed'],
-                        $mapping['aggregate']
-                    )
-                );
+                if ($isBlockScope && $value && Field::TYPE_STRING === $type) {
+                    $this->blockValues[] = \strip_tags($value);
```
alexander-schranz: strip_tags shouldn't be handled here. And we should not use a property to fill blockValues as written above services should be always stateless.

Search/ObjectToDocumentConverter.php
```diff
@@ -110,8 +115,16 @@ public function objectToDocument(IndexMetadata $metadata, $object)
             $document->setLocale($locale);
         }
 
+        $this->blockValues = [];
```
alexander-schranz: As written above this should just be a local variable and not a class property. To keep service stateless.

alexander-schranz (changes requested):

Search/ObjectToDocumentConverter.php
alexander-schranz: I'm not sure how we did handle text editors yet if there is such a transformer we need still use that one. strip_tags should not be used on none text editor fields. Else it will drop a title like `Test <Hello some more text which is dropped`.

So not 100% sure if as spoken we still need a "ContentMetadata" which is the merged metadata to keep things clearer.

/cc @wachterjohannes do you have here some insights :)

thomasduenser: @alexander-schranz @wachterjohannes Until yet, all the HTML tags were kept in the index. If that's what you prefer, it's fine for me too!

If I've seen it right, the "ComplexMetadata" is only used for block-related stuff.

Search/ObjectToDocumentConverter.php
thomasduenser: @alexander-schranz That's fine, should I use Pass by Reference and modify the `$blockValues` variable or should I return the `$blockValues` variable in the `populateDocument` method instead?

Search/ObjectToDocumentConverter.php
alexander-schranz: Okay I thought snippets, categories, were also complex metadata but was not sure. Lets see what @wachterjohannes think about it.

Search/ObjectToDocumentConverter.php
```diff
@@ -110,8 +115,16 @@ public function objectToDocument(IndexMetadata $metadata, $object)
...
         $this->populateDocument($document, $object, $fieldMapping);
 
+        // Adds the merged data of each content-block (even nested-blocks) to the document.
+        if (0 < \count($this->blockValues)) {
+            $mapping = $this->addMappingOptions();
+            $blockValues = \implode(' ', $this->blockValues);
+            $this->addDocumentField($document, 'contentBlocks', $blockValues, $mapping, Field::TYPE_STRING);
```
wachterjohannes: @thomasduenser we should use a field for that - currently we something like that:

```
blocks0title = "Title 1"
blocks1title = "Title 2"
```

I think in the future we should have:

```
blocks_title = "Title 1,Title 2"
```

And that should be done in the field mapping and not here in the converter.

So i think we should change the behaviour of the `ComplexMetatada` to not populate each value as a own field in the index.

wachterjohannes: @alexander-schranz what do you think about the behaviour change?

Search/ObjectToDocumentConverter.php
wachterjohannes: @thomasduenser i would no change the behaviour here and strip the html tags - if this was the case before i would keep it and add an issue to this repo. Then we could discuss that and tackle that by its own.

wachterjohannes: for the rest i have added a comment above

wachterjohannes: @thomasduenser as an addition the generation of the complex metadata happens here: https://github.com/sulu/sulu/blob/2.5/src/Sulu/Bundle/PageBundle/Search/Metadata/StructureProvider.php#L312-L320

Search/ObjectToDocumentConverter.php
alexander-schranz: @wachterjohannes i would do that I would store all into one field aslong as they have the same settings. If we support boost we would create a field per boost level.

wachterjohannes: @alexander-schranz but i think that is a different issue - we could store all the data of the content in an own field! but that should be a configuration - should this field be an own field or not (could be NOT as default).

But to fix this issue to map block properties (with own field like the current behaviour is) there should be an own field for each `sulu.search.field` which could be used to boost for example. But also for that we would have to combine all the values.

alexander-schranz: @wachterjohannes and that is why I currently would just create one "content" field to store all fields into it instead of create seperate fields.

wachterjohannes: @alexander-schranz as i said before :) should be configureable by field - but currently i would just fix the block issue - everything else should be done i a second PR after digging into it!

thomasduenser: @wachterjohannes @alexander-schranz Then I'll fix the issues in this PR (stateless service, remove strip_tags)? And the other improvements (configuration per field) will be discussed/changed at another time?

Search/ObjectToDocumentConverter.php
wachterjohannes: i would do the change described in the comment https://github.com/massiveart/MassiveSearchBundle/pull/168#discussion_r1002912687 in this PR and create an issue for a "overall" content field

wachterjohannes: @thomasduenser @alexander-schranz that would fix the current issue and it does not change the behaviour to much

alexander-schranz: @wachterjohannes it is easier to store all into one field. Why split it into several fields? I see no advantages or usage of several fields here.

thomasduenser: @wachterjohannes @alexander-schranz We have many projects with a website search and never used a block field directly. It is fine to store all the values in one field (to be searchable) without the option to configure anything.

For me the generation of the complex metadata looks quite complex and the code parts in this bundle are way more understandable.

The changes in this PR would fix our struggles and that's why I'd also prefer the easy way right now.

wachterjohannes: @alexander-schranz i dont like to mix that issues here - it feels to me more like a hack than a quick fix

wachterjohannes: @thomasduenser yeah thats possible that you dont use it as a search field or something else but i think we should not change this behaviour here when it is no neccesary.

My suggested changes should also be possible to fix without changing the metadata creation in sulu. The changes would be also in the search-bundle  itself

thomasduenser: @wachterjohannes I can try to implement your wanted changes here in this bundle as you described [here](https://github.com/massiveart/MassiveSearchBundle/pull/168#discussion_r1002912687).

Does that mean that I can ignore this?
> And that should be done in the field mapping and not here in the converter.

I'll start to change the Pull Request once you and @alexander-schranz have agreed on a way forward.

wachterjohannes: @thomasduenser yeah you can ignore that sentence :) that was a bit misunderstandable :)

wachterjohannes (approved):
@alexander-schranz what do you think?

PR #172, opened 2023-09-18
Allow Symfony 7 versions of packages

alexander-schranz:
Requires:

 - [x] matthiasnoback/symfony-dependency-injection-test Symfony 7 support


Command/InitCommand.php
```diff
@@ -41,7 +43,7 @@ public function configure()
         $this->setHelp('This command will simply call the initialize method of the currently active search adapter.');
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output)
+    protected function execute(InputInterface $input, OutputInterface $output): int
```
alexander-schranz, line 46: small BC Break in all commands, but think the commands execute method should not be overridden.

Prokyonn (approved):

PR #175, opened 2024-02-08
Fix MassiveSearchExpressionLanguage with object/stdClass for block settings

popoplanter:
Fixes search reindex for ArticleDocument and PageDocument by catching NoSuchIndexException


alexander-schranz:
Settings when empty are a `\stdClass` this currently required because how JSON Serializer works as empty array is converted to `[]` but the frontend expect an object `{}` for the Sulu JADMIN and so it is `\stdClass` which seems to fail currently in the massive search bundle property accessor. The `NoSuchIndexException` is throw when tried access a object by array selectors. This way we avoid currently this error and return the default value. /cc @wachterjohannes

wachterjohannes (approved):

wachterjohannes:
@popoplanter the change seems legit! but please fix php-cs

PR #176, opened 2024-03-20
Fix SearchEvent proper return type hint

sauls:
I am using phpstan with strict rules and max level and this caught this slipped wrong type hint.


alexander-schranz (approved):

alexander-schranz:
@sauls Thank you!

PR #177, opened 2024-04-19
Use search manager interface instead of actual class

fnagel:
Reindex command is the only command using an actual class instead of the interface: This causes issue when trying to decorate the search manager.


alexander-schranz:
@fnagel Thank you. Released in 2.9.1

PR #179, opened 2024-12-02
Fix PHP 8.4 deprecations nullable objects

alexander-schranz:


composer.json
```diff
@@ -32,7 +32,7 @@
         "symfony/security-bundle": "^4.3 || ^5.0 || ^6.0 || ^7.0",
         "symfony/twig-bundle": "^4.3 || ^5.0 || ^6.0 || ^7.0",
         "symfony/validator": "^4.3 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/phpunit-bridge": "^5.0.4 || ^6.0 || ^7.0",
+        "symfony/phpunit-bridge": "^5.0.4 || ^6.0 || 7.1.*",
```
alexander-schranz: The 7.2 bridge currently fails with: PHP Fatal error:  Uncaught Error: Call to private PHPUnit\Framework\TestSuite::__construct() from scope PHPUnit\Util\Configuration in /home/runner/work/MassiveSearchBundle/MassiveSearchBundle/vendor/bin/.phpunit/phpunit-8-1/src/Util/Configuration.php:990

alexander-schranz: Looks like this is failing since a longer time without a proper fix yet. So I will ignore the issue.

Prokyonn (approved):

PR #180, opened 2025-01-03
Added explicit phpdoc for return type of bundle's build method

stollr:
This is needed to prevent the following deprecation message in a backwards compatible way:

> Method 'Symfony\Component\HttpKernel\Bundle\Bundle::build()' might add 'void' as a native return type declaration in the future. Do the same in child class 'Massive\Bundle\SearchBundle\MassiveSearchBundle' now to avoid errors or add an explicit @return annotation to suppress this message. in vendor/symfony/error-handler/DebugClassLoader.php on line 341

Question: What is the preferred way to add a native return type?


alexander-schranz:
@stollr thank you

PR #184, opened 2025-04-08
Add check for non array in field-evaluator

wachterjohannes:
The `$this->expressionLanguage->evaluate($condition, $object);` expect only a array.

### DIFF
--- a/Search/Metadata/FieldEvaluator.php
+++ b/Search/Metadata/FieldEvaluator.php
@@ -72,6 +72,10 @@ class FieldEvaluator
 
     public function evaluateCondition($object, string $condition)
     {
+        if (!\is_array($object)) {
+            return false;
+        }
+
         try {
             return $this->expressionLanguage->evaluate($condition, $object);
         } catch (\Exception $e) {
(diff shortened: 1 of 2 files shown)


Tests/Unit/Search/Metadata/FieldEvaluatorTest.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 /*
```
alexander-schranz:
```suggestion
/*
```

Search/Metadata/FieldEvaluator.php
```diff
@@ -72,6 +72,10 @@ public function getValue($object, FieldInterface $field)
 
     public function evaluateCondition($object, string $condition)
     {
+        if (!\is_array($object)) {
```
alexander-schranz, line 75: why the object should be an array and not an object? what is the issue you try to solve here?

The variable is called object so I think it should handle also objects not only arrays

wachterjohannes: the function can be called with any type and the expression language requires an array - so returning false

PR #186, opened 2025-10-30
Avoid serialization loop with IndexMetadata during serialization

alexander-schranz:
Issues appear during saving a category or similiar in sulu admin the metadata serilaization ending in:

> terminated by signal SIGSEGV (Address boundary erro

Or for others as reported

> [2025-10-29T15:49:24.795061+01:00] request.CRITICAL: Uncaught PHP Exception Error: "Maximum call stack size of 8339456 bytes (zend.max_allowed_stack_size - zend.reserved_stack_size) reached. Infinite recursion?" at SerializationHelper.php line 16 {"exception":"[object] (Error(code: 0): Maximum call stack size of 8339456 bytes (zend.max_allowed_stack_size - zend.reserved_stack_size) reached. Infinite recursion? at /data/sites/web/<user>/subsites/<subsite>/vendor/jms/metadata/src/SerializationHelper.php:16)"} []

---

PS: Issues is related to the newly released versions of PHP 8.3.27 and 8.4.14, see https://github.com/php/php-src/issues/20338, PHP versions before are not effect by the bug, but update still recommended.

### DIFF
--- a/.github/workflows/test-application.yaml
+++ b/.github/workflows/test-application.yaml
@@ -78,7 +78,15 @@ jobs:
                     - php-version: '8.3'
                       elasticsearch-version: '7.17.19'
                       elasticsearch-package-constraint: '~7.17.0'
-                      minimum-stability: 'dev'
+                      minimum-stability: 'stable'
+                      dependency-versions: 'highest'
+                      tools: 'composer:v2'
+                      php-cs-fixer: false
+
+                    - php-version: '8.4'
+                      elasticsearch-version: '7.17.19'
+                      elasticsearch-package-constraint: '~7.17.0'
+                      minimum-stability: 'stable'
                       dependency-versions: 'highest'
                       tools: 'composer:v2'
                       php-cs-fixer: false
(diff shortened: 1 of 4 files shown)


martinlagler (approved):

matthiasnoback:
Hi @alexander-schranz Thanks for fixing this bug. It's really strange, because we've been using Sulu with the Search functionality for some time now and suddenly this bug appears. We don't use Categories. Doing a data dump inside the infinite loop it it serializing metadata for Account. However, in terms of data or metadata nothing has changed, so I can't explain this issue suddely appearing. Do you have any clues?

alexander-schranz:
Hi @matthiasnoback, I think it happens during creating the metadata, so I think even if you don't use the categories.

Suddenly appears in this case means you did get newer versions of PHP on your System. By maybe using `:8.3` and `:8.4` docker tags or apt-get updated your dependency. There were changes in PHP core itself (8.3.27 and 8.4.14) which make this bug appear.

If you run in all versions https://3v4l.org/40nda you see it worked well before 8.3.27 and 8.4.14 but fails on the new one. I created an issue there: https://github.com/php/php-src/issues/20338. But not sure how they will handle it. For us it needed to be a quickfix as it crashed several PHP installation so we avoided the index loop.

To be sure, updating the MassiveSearchBundle did fix it also in your project?

matthiasnoback:
Thanks, @alexander-schranz ! Of course, implicitly updating PHP 8.3 while building the Docker container can explain this. Indeed, updating the bundle fixed it.

PR #190, opened 2026-02-10
Drop support for PHP <= 8.1 and Symfony <= 5.3

alexander-schranz:
We not longer support this versions in Sulu and keep maintainance low and CI easier it best to drop this.


.github/workflows/test-application.yaml
alexander-schranz: not sure about this 🤔 -> elastic not tests on > 8.0 currently

composer.json
```diff
@@ -10,34 +10,34 @@
...
+        "symfony/monolog-bridge": "^5.4 || ^6.0 || ^7.0",
         "symfony/monolog-bundle": "^3.1",
-        "symfony/security-bundle": "^4.3 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/twig-bundle": "^4.3 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/validator": "^4.3 || ^5.0 || ^6.0 || ^7.0",
-        "symfony/phpunit-bridge": "^5.0.4 || ^6.0 || ^7.0 ",
+        "symfony/security-bundle": "^5.4 || ^6.0 || ^7.0",
+        "symfony/twig-bundle": "^5.4 || ^6.0 || ^7.0",
+        "symfony/validator": "^5.4 || ^6.0 || ^7.0",
+        "symfony/phpunit-bridge": "^6.4 || ^7.0 ",
```
Prokyonn, line 35: 5.4 not possible?

alexander-schranz: did fail on newer phpunit version, But it does not matter as symfony/phpunit-bridge does not have any requirements only php >=7.1

Prokyonn (approved):

